PR #28, opened 2015-03-02
FEATURE(sla): moves sla-core into platform

rosogon:
Integrating the sla-core project into the platform.


adriannieto:
@rosogon Can you review what's failing in the build process? Please take a look to https://buildhive.cloudbees.com/job/SeaCloudsEU/job/SeaCloudsPlatform/105/consoleFull and try to build by yourself the whole project to realize what's happening. 

Thank you!

rosogon:
I use a configuration.properties to customize the compiling. This file is not versioned, but a configuration.properties.sample, that should be copied to configuration.properties. 

Of course, travis is not able to handle that. If you don't any better idea, I'll git add configuration.properties and remove the sample file.

adriannieto:
I don't know how to deal with this, maybe @andreaturli  can point us in the right direction.

mbarrientos:
@rosogon Another option could be to use [maven resources plugin](http://maven.apache.org/plugins/maven-resources-plugin/) to generate the properties file for the build (and maybe removing it using  'clean' task on that specific file afterwards).

Also, if there are some integration tests failing i'd recommend you to mark them so ([pom.xml](https://github.com/SeaCloudsEU/SeaCloudsPlatform/blob/master/pom.xml#L88)), since they're excluded from the default test phase.

rosogon:
@mbarrientos 
I don't think maven resources plugin can do what I need: copy and rename. There is a workaround with maven assembly plugin, but I don't like this idea.

The best idea may be something like this: http://stackoverflow.com/a/11366713

adriannieto:
@rosogon please squash the commits into the first one, and we will be ready to merge it, thank you very much.

rosogon:
Commits squashed into one && git push -f

PR #31, opened 2015-04-23
FIX/DOCS: Update CONTRIBUTING.md

andreaturli:
@mbarrientos can you have a look and merge it, if ok?


mbarrientos:
It's OK for me. Looks much more solid now. Thanks!

PR #32, opened 2015-04-23
Implementing SeaClouds monitoring platform as a RestFul Web Service.

dionysiosAthanasopoulos:


andreaturli:
@dionysiosAthanasopoulos thanks for the PR.

I think 763 files is definitely too much for one single commit, though!

Also looking at the files committed, they seem to be mostly dependencies for your code, not really something you have developed (lots of third-party jars, etc) 

I think it is best to restructure your project  as a maven project and define there all the dependencies, instead of committing them in the repo. wdyt?

andreaturli:
I think that might be useful in this case http://stackoverflow.com/questions/11731634/problems-using-sigar-with-maven2

mbarrientos:
Hi @dionysiosAthanasopoulos. I see you're having some problems with the build, some general comments about it:
- Try to check locally first that your code is building properly before pushing changes. This is done by running `mvn clean install` on the root folder. If it builds with SUCCESS then you might want to push the commits.
- TravisCI build might fails on `deploy` phase, like it happened the last build, but that's a known issue and it doesn't have anything to do with your code. Don't give much importance to that now.
- It's also a nice practice to squash your commits into one since you're updating the PR and not adding new content. This might be useful: https://github.com/edx/edx-platform/wiki/How-to-Rebase-a-Pull-Request#squash-your-changes-optional

mbarrientos:
This PR has now moved to https://github.com/SeaCloudsEU/SeaCloudsPlatform/pull/41. @dionysiosAthanasopoulos can we close it now?

PR #33, opened 2015-04-24
FIX(DASHBOARD): Dashboard connection with deployer is now wired through Java properties.

mbarrientos:
Dashboard was not able to be configured dynamically. Now it expects the following properties to connect to the Deployer engine:
- deployer.host
- deployer.httpPort
- deployer.username
- deployer.password

Also (provisionally, not being used now) supporting properties for the rest of modules:
- planner.endpoint
- monitor.endpoint
- sla.endpoint

### DIFF
--- a/dashboard/src/main/java/eu/seaclouds/platform/dashboard/ConfigParameters.java
+++ b/dashboard/src/main/java/eu/seaclouds/platform/dashboard/ConfigParameters.java
@@ -16,8 +16,46 @@
  */
 package eu.seaclouds.platform.dashboard;
 
-public interface ConfigParameters {
-    public static String MONITOR_ENDPOINT = "http://localhost:8081";    // CHANGE THIS
-    public static String DEPLOYER_ENDPOINT = "http://localhost:8081";   // CHANGE THIS
-    public static String SLA_ENDPOINT = "http://localhost:8082";        // CHANGE THIS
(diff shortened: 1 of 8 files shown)


dashboard/src/main/java/eu/seaclouds/platform/dashboard/ConfigParameters.java
```diff
@@ -16,8 +16,46 @@
  */
 package eu.seaclouds.platform.dashboard;
 
-public interface ConfigParameters {
-    public static String MONITOR_ENDPOINT = "http://localhost:8081";    // CHANGE THIS
```
andreaturli: so the problem with the dashboard was that, not tomcat vs jetty, right?

andreaturli:
looks better now. We need to update SeaCloudsEU/seaclouds-distribution/seaclouds.yaml accordingly

andreaturli:
thanks @mbarrientos, merging now

dashboard/src/main/java/eu/seaclouds/platform/dashboard/ConfigParameters.java
mbarrientos: It was both issues. Tomcat was not loading the servlets properly, probably because of some dependency issue, but not sure. However I've been testing it with Jetty without any problem.

PR #39, opened 2015-04-27
Optimizer second version

perezp:
Main improvements of this version: Optimizer considers network latency in the performance evaluation of a candidate deployment. Unification in the utilization of time units (millis, seconds, hours) and availability metrics (percentage and proportion). Repair as many bugs as they were identified in the meantime.

I've tried to squash the commits into only one, with date April 26th (day of the pull request). I don't know if I have succeeded on it… let me know.

@andreaturli @mbarrientos @kiuby88 Could you give me your opinion of the PR?


planner/optimizer/src/main/java/eu/seaclouds/platform/planner/optimizer/CloudOffer.java
```diff
@@ -91,7 +97,8 @@ public void setCost(double cost) {
 
    @Override
    public CloudOffer clone() {
```
adriannieto: In my opinion, in order to clone objects in Java maybe is better to use the native interface to perform this (http://stackoverflow.com/questions/869033/how-do-i-copy-an-object-in-java) also watch out with the member variables if any of them are references to objects (like lists) and you don't clone the list but the reference (it works with strings because in Java strings are immutable, but just in case take this into consideration)

adriannieto:
Regarding to the output files:

planner/optimizer/src/test/java/eu/seaclouds/platform/planner/optimizerTest/resources/outputHILLCLIMB0.yaml
planner/optimizer/src/test/java/eu/seaclouds/platform/planner/optimizerTest/resources/outputHILLCLIMB1.yaml
planner/optimizer/src/test/java/eu/seaclouds/platform/planner/optimizerTest/resources/outputHILLCLIMB2.yaml
planner/optimizer/src/test/java/eu/seaclouds/platform/planner/optimizerTest/resources/outputHILLCLIMB3.yaml
planner/optimizer/src/test/java/eu/seaclouds/platform/planner/optimizerTest/resources/outputHILLCLIMB4.yaml

Are they generated with the tests? (If the response is yes, they should be added to .gitignore file)

adriannieto:
I'm running the tests in my computer and I got this, can you check if am I missing something?   (I just downloaded your branch and performed a mvn clean install)

```
-------------------------------------------------------------------------------
Test set: TestSuite
-------------------------------------------------------------------------------
Tests run: 7, Failures: 2, Errors: 0, Skipped: 0, Time elapsed: 2.16 sec <<< FAILURE!
testPresenceSolutionBlind(eu.seaclouds.platform.planner.optimizerTest.OptimizerTest)  Time elapsed: 0.311 sec  <<< FAILURE!
java.lang.AssertionError: Dam was not created
	at org.junit.Assert.fail(Assert.java:88)
	at org.junit.Assert.assertTrue(Assert.java:41)
	at eu.seaclouds.platform.planner.optimizerTest.OptimizerTest.checkCorrectness(OptimizerTest.java:143)
	at eu.seaclouds.platform.planner.optimizerTest.OptimizerTest.testPresenceSolutionBlind(OptimizerTest.java:96)

testPresenceSolutionHillClimb(eu.seaclouds.platform.planner.optimizerTest.OptimizerTest)  Time elapsed: 0.298 sec  <<< FAILURE!
java.lang.AssertionError: Dam was not created
	at org.junit.Assert.fail(Assert.java:88)
	at org.junit.Assert.assertTrue(Assert.java:41)
	at eu.seaclouds.platform.planner.optimizerTest.OptimizerTest.checkCorrectness(OptimizerTest.java:143)
	at eu.seaclouds.platform.planner.optimizerTest.OptimizerTest.testPresenceSolutionHillClimb(OptimizerTest.java:122)
```

planner/optimizer/src/main/java/eu/seaclouds/platform/planner/optimizer/CloudOffer.java
perezp: I completely agree with you. There is the opinion that Java clone() is broken and should not be used. 
I took the issue of the objects references into account. In fact, the clone() method in class SuitableOptions (in this same package) cares about doing the deep clone. Actually, this clone() is only called by the clone in SuitableOptions class, while doing the deep copy. If you follow the opinion that clone() method is broken and we must avoid its utilization, I'll create a constructor that receives as parameter an object of the same class. 

Let me know if your comment was an opinion to discuss or a command to me to change it ;) .

perezp:
@adriannieto 
"Regarding to the output files: … "

You're right. Indeed it is written in the .gitignore!!
I think the problem was that, in order to make the PR with a single commit, I pasted ALL the files changed at once and I did the "git add --all" and "git commit" …. Now the funny part: among the "ALL" the files changed, it was the .gitignore (which was changed to avoid including the output files). So, during the general "add", it could not take into account the .gitignore because this file was also being added in the same command (facepalm). It will not happen again (hopefully). Can I remove these files directly from here?

Thanks!!

planner/optimizer/src/main/java/eu/seaclouds/platform/planner/optimizer/CloudOffer.java
adriannieto: It was an opinion, if it's working no problem with this. I just wanted to notice you.

adriannieto:
@perezp well about the files don't worry, it's a minor issue. You cannot remove them in the web interface.  I'm more concerned about the failing tests thing.

perezp:
The tests on windows were failing due the line ending. Fixed by modifing the options on the SnakeYaml parser  . Thanks @adriannieto

adriannieto:
It looks OK for me, merging it.

PR #40, opened 2015-04-27
FIX(Dashboard): fixing succesive connections problem.

mbarrientos:
Fixes bug where successive requests to servlets were not re-allocating a new connection to the Brooklyn API, hence throwing a exception.


adriannieto:
It will be nice to encapsulate all the Brooklyn calls inside the Deloyer API, but this is outside the scope of this PR. It looks OK for me.

PR #41, opened 2015-04-27
Monitor RestFul Service v1

dionysiosAthanasopoulos:


pom.xml
```diff
@@ -134,7 +139,7 @@
           <target>1.7</target>
         </configuration>
       </plugin>
-      <plugin>
+      <!--<plugin>
```
mbarrientos: Why are you commenting this? This plugin should check the existance of license in every source file.

monitor/pom.xml
```diff
@@ -27,36 +27,50 @@
...
       <groupId>org.apache.brooklyn</groupId>
       <artifactId>brooklyn-rest-client</artifactId>
-      <version>0.7.0-SNAPSHOT</version>
-    </dependency>
+    </dependency> 
+    <!--<dependency>
+    	<groupId>io.brooklyn</groupId>
+    	<artifactId>brooklyn-rest-client</artifactId>
+    	<version>0.7.0-M1</version>
+	</dependency>-->
```
mbarrientos: Please, remove commented code.

monitor/api/pom.xml
```diff
@@ -20,13 +20,320 @@
...
+		</repository>
+
+		<repository>
+			<id>deib-polimi-snapshots</id>
+			<url>https://github.com/deib-polimi/deib-polimi-mvn-repo/raw/master/snapshots</url>
+		</repository>
+
+		<repository>
+        	<id>imperial-modaclouds-releases</id>
+        	<url>https://github.com/imperial-modaclouds/modaclouds-data-collectors/releases/download/1.3-Snapshot</url>
```
mbarrientos: The link is broken (maybe https://github.com/imperial-modaclouds/modaclouds-data-collectors/releases/download/1.3-Snapshot/data-collector-1.3-SNAPSHOT.jar ?). However, I don't think single URL's to files can work as maven repositories.

monitor/api/pom.xml
```diff
@@ -20,13 +20,320 @@
...
+                            	<include>**/data-collector-1.3-SNAPSHOT.jar</include>
+                            	<include>**/dataCollectorsInstallation.bat</include>
+                        	</includes>
+                        	<followSymlinks>false</followSymlinks>
+                    	</fileset>
+                	</filesets>
+            	</configuration>
+        	</plugin>
+
+			<!--<plugin>
```
mbarrientos: Please, remove commented code.

monitor/api/src/main/java/resources/Monitor.java
```diff
@@ -0,0 +1,874 @@
...
+
+		return "[INFO] Monitor REST Service: Initiating monitor...\n" + msg;
+	}
+
+	@POST
+	@Produces("text/plain")
+	@Path("/initiate")//http://localhost:8080/monitor-api/rest/initiate/xxxxxx
+	public String initiate( String serverSourcesPath ) {
+
+		//serverSourcesPath = serverSourcesPath.replace( '+', '/' ); //System.out.println( "\nserverSourcesPath = " + serverSourcesPath );
```
mbarrientos: Same here, commented code.

monitor/api/src/main/java/resources/Monitor.java
```diff
@@ -0,0 +1,874 @@
...
+				ex.printStackTrace();
+			}
+		}
+
+
+		return msg;
+	}
+
+
+	public static void main( String[] args ) {
```
mbarrientos: This main method looks like it serves a different purpose than the rest of this source file. Would it make sense to keep it in a different file, something like MonitorCLI?

monitor/core/src/main/java/core/ControllerImpl.java
```diff
@@ -0,0 +1,417 @@
...
+		if( PlannerURIRulesReady != null ) fileContent += "\nPlannerURIRulesReady=" + PlannerURIRulesReady;
+		if( PlannerURIReplanning != null ) fileContent += "\nPlannerURIReplanning=" + PlannerURIReplanning;
+
+
+		new File( SEACLOUDS_FOLDER ).mkdirs();
+
+		TxtFileWriter.write( fileContent, INITIALIZATION_CONFIGURATION_FILE_ON_SERVER );
+	}
+
+	/*private void storeInitializationData( InitializationData initializationData ) throws IOException{
```
mbarrientos: Commented code.

monitor/core/src/main/java/core/ControllerImpl.java
```diff
@@ -0,0 +1,417 @@
...
+		ControllerImpl controller = new ControllerImpl();
+
+		controller.initiateMonitor();
+
+
+		try{
+
+			System.out.println( "\nSeaClouds Monitoring Platform Menu:" );
+
+			while( true ){
```
mbarrientos: Please remove this loop.

mbarrientos: In case that you wanted to keep a similar CLI as Monitor.java, I'll recommend to uncoment necessary code and keep it separatedly.

mbarrientos:
Thanks for creating the PR @dionysiosAthanasopoulos. 

As a general guideline:
- try to remove commented and/or unnecesary code out of the PR, to make it simpler and better to assimilate.
- Also you need to add license headers in every source file (mostly Java and XML) to make the rat check pass (you can have a look on the license of any other source file in the repository).
- I'm not sure on what to do about the uploaded jar libraries. I see that those libraries are publicly available but not on a maven repository. @andreaturli, do you have any idea/experience on how to treat this? 
- Additionally (also optionally), I'd like to see some tests to illustrate the functionality on this monitor, but as I say, maybe is not necessary for this first PR.

monitor/core/src/main/java/core/JsonParsing.java
```diff
@@ -0,0 +1,131 @@
...
+
+/**
+ * 
+ * @author Dionysis Athanasopoulos <dionysiscsuoi@gmail.com>
+ *
+ */
+
+public class JsonParsing {
+
+	private static void getArray( Object object ) throws ParseException {
```
adriannieto: An static method which doesn't return anything and even with no static variables in this class must be removed.

rosogon: It's "returning" allArrayElements.

adriannieto: Where is the return? Where is the return type?

rosogon: The code modifies the list passed as parameter. It will be clearer if the function returns the list, but the function CAN'T be removed.

adriannieto: In my opinion, this kind of things difficult a lot the understanding of the code, if he want to return a list the proper way to do it is to initialize inside of the method, or returning a copy of the list.

andreaturli: there are tons of json parser out there, gson, jackson just to name the most famous. Do we really want to maintain this custom code?

monitor/api/pom.xml
```diff
@@ -20,13 +20,320 @@
...
+            			<goals>
+              				<goal>run</goal>
+            			</goals>
+          			</execution>
+        		</executions>
+      		</plugin>
+
+			<plugin>
+    			<artifactId>maven-antrun-plugin</artifactId>
+    			<version>1.6</version>
```
andreaturli: why is it needed? I thinl `antrun` shouldn't be needed. wdyt @dionysiosAthanasopoulos ? thanks

monitor/api/src/main/java/resources/Monitor.java
```diff
@@ -0,0 +1,874 @@
...
+import core.ControllerImpl;
+import core.TxtFileReader;
+import core.TxtFileWriter;
+import core.WindowsBatchFileExecution;
+import core.RESTCalls.RESTGet;
+import core.RESTCalls.RESTPost;
+
+/**
+ * 
+ * @author Dionysis Athanasopoulos <dionysiscsuoi@gmail.com>
```
andreaturli: we are removing the @author tag from the code.

monitor/core/pom.xml
```diff
@@ -20,15 +20,181 @@
...
+                         			usetimestamp="true"/>
+                         		<!--<get src="http://www.cs.uoi.gr/~dathanas/monitoring-manager-1.4.jar"
+                         			dest="${project.build.directory}"
+                         			verbose="false"
+                         			usetimestamp="true"/>
+                         		<get src="https://github.com/imperial-modaclouds/modaclouds-data-collectors/releases/download/1.3-Snapshot/data-collector-1.3-SNAPSHOT.jar"
+                         			dest="${project.build.directory}"
+                         			verbose="false"
+                         			usetimestamp="true"/>-->
+                 			</target>
```
mbarrientos: These commented targets look like an alternative (not the best, but i'd be ok to merge with this at the moment) to having to add the jars to the repository. Is it commented for any reason @dionysiosAthanasopoulos ?

monitor/core/pom.xml
```diff
@@ -20,15 +20,181 @@
...
+                         			verbose="false"
+                         			usetimestamp="true"/>
+                    			<get src="https://magelan.googlecode.com/files/hyperic-sigar-1.6.4.zip"
+	        			            dest="${project.build.directory}"
+    	                		    verbose="false"
+        	                 		usetimestamp="true"/>
+                    			<get src="http://www.cs.uoi.gr/~dathanas/rsp-services-csparql-0.4.6.2-modaclouds-distribution.zip"
+                         			dest="${project.build.directory}"
+                         			verbose="false"
+                         			usetimestamp="true"/>
```
mbarrientos: These 3 ant artifacts apparently can be found in maven central repo:
- http://mvnrepository.com/artifact/org.apache.jena/jena-fuseki/1.1.1
- http://mvnrepository.com/artifact/org.fusesource/sigar/1.6.4
- http://streamreasoning.org/maven/polimi/deib/rsp-services-csparql/0.4.6/

I'm not sure about if the hyperic-sigar is the same artifact. Could you check that they are suitable for the project and replace them as Maven dependency? Thanks.

monitor/pom.xml
```diff
@@ -71,6 +73,51 @@
           <target>1.7</target>
         </configuration>
       </plugin>
+
+			<plugin>
+        		<artifactId>maven-antrun-plugin</artifactId>
+        		<version>1.6</version>
+        		<executions>
+          			<execution>
+            			<phase>clean</phase>
```
mbarrientos: This task should not be linked to maven `clean` phase. Please change this linking (and the one just below) to `generate-sources`.

monitor/api/pom.xml
mbarrientos: I would only save the `antrun` tasks for the 2 conflicting libraries (see https://github.com/SeaCloudsEU/SeaCloudsPlatform/pull/41/files#diff-344e50cefa0780cf204498bf7450e6ecR78)  which don't have Maven support (as a temporal solution). I have the impression that the rest of the libraries being downloaded can be found in a public Maven repository. Please convert them to Maven dependencies if possible.

adriannieto:
I have a remark about the coding style, there are a lot of unnecessary line breaks (among with some indentations problems), it should be fixed before merging it, you may use a tool to review if your follow the code convention http://checkstyle.sourceforge.net/

rosogon:
I saw indentation problems due to mixing spaces and tabs. If you use eclipse, there is an option to show tabs.

adriannieto:
You're right, In order to avoid problems all tabs should be replaced by whitespaces 
El 30/4/2015 4:27 p. m., Román SG <notifications@github.com> escribió:I saw indentation problems due to mixing spaces and tabs. If you use eclipse, there is an option to show tabs.

adriannieto:
Please could you provide some tests to expose how this module works? (You can see an example of this in https://github.com/SeaCloudsEU/SeaCloudsPlatform/pull/39), it will ease the integration of this PR and the further work.

monitor/api/src/test/java/monitorTest/MonitorTest.java
```diff
@@ -0,0 +1,321 @@
...
+	private static final String DEPLOYMENT_MODEL_FILE = "/api/resources/chat-WebApplication.json";
+	private static final String DATA_COLLECTORS_FILE_NAME = "data-collector-1.3-SNAPSHOT.jar";
+	private static final String DATA_COLLECTORS_INSTALLATION_FILE_NAME = "dataCollectorsInstallation.bat";
+
+	static Logger log = LoggerFactory.getLogger(MonitorTest.class);
+
+	@Test
+	public void test() {
+
+		if (!OperatingSystem.isWindows())
```
adriannieto: This test will finish successfully in any other OS different than windows (this is why the tests passes in GH automatically). @dionysiosAthanasopoulos,  does the monitor only work on Windows? If the test refers to a single platform maybe it will be less confusing to name the tests MonitorWindowsTest or something like this. If your software is multiplatform please provide some tests for unix/osx (I think that the incompatibility is that you're using windows batch files to initialize ModaClouds, it should be possible to do the same task by using bash scripting)

Anyway as we need to integrate as soon as possible this PR with the master branch, I will +1 it if you think it's necessary, giving you more time to fix those problems for the next release of the Monitor. 

Thank you very much.

adriannieto:
Although there are many open issues, we need to integrate this now. They will be taken into account in the next version. Merging it.

PR #42, opened 2015-05-07
Planner

szenzaro:
Pull request for matchmaker and planner

### DIFF
--- a/planner/core/pom.xml
+++ b/planner/core/pom.xml
@@ -20,12 +20,13 @@ limitations under the License.
          xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
     <modelVersion>4.0.0</modelVersion>
     <groupId>eu.seaclouds-project</groupId>
-    <artifactId>core</artifactId>
-    <version>1.0-SNAPSHOT</version>
+    <artifactId>planner-core</artifactId>
+    <version>0.1.1-SNAPSHOT</version>
+    <!--packaging>war</packaging-->
     <parent>
         <groupId>eu.seaclouds-project</groupId>
         <artifactId>planner</artifactId>
-        <version>1.0-SNAPSHOT</version>
+        <version>0.1.1-SNAPSHOT</version>
     </parent>
     <dependencies>
         <dependency>
--- a/planner/matchmaker/pom.xml
+++ b/planner/matchmaker/pom.xml
@@ -23,11 +23,11 @@ limitations under the License.
   <parent>
     <groupId>eu.seaclouds-project</groupId>
     <artifactId>planner</artifactId>
-    <version>1.0-SNAPSHOT</version>
+    <version>0.1.1-SNAPSHOT</version>
   </parent>
   <groupId>eu.seaclouds-project</groupId>
   <artifactId>matchmaker</artifactId>
-  <version>1.0-SNAPSHOT</version>
+  <version>0.1-SNAPSHOT</version>
   <dependencies>
     <dependency>
       <groupId>eu.seaclouds-project</groupId>
--- a/planner/discoverer/pom.xml
+++ b/planner/discoverer/pom.xml
@@ -23,11 +23,11 @@ limitations under the License.
     <parent>
         <groupId>eu.seaclouds-project</groupId>
         <artifactId>planner</artifactId>
-        <version>1.0-SNAPSHOT</version>
+        <version>0.1.1-SNAPSHOT</version>
     </parent>
     <groupId>eu.seaclouds-project</groupId>
     <artifactId>discoverer</artifactId>
-    <version>1.0-SNAPSHOT</version>
+    <version>0.0.1-SNAPSHOT</version>
     <dependencies>
         <dependency>
             <groupId>eu.seaclouds-project</groupId>
--- a/planner/pom.xml
+++ b/planner/pom.xml
@@ -21,7 +21,7 @@ limitations under the License.
     <modelVersion>4.0.0</modelVersion>
     <groupId>eu.seaclouds-project</groupId>
     <artifactId>planner</artifactId>
-    <version>1.0-SNAPSHOT</version>
+    <version>0.1.1-SNAPSHOT</version>
 
     <parent>
         <groupId>eu.seaclouds-project</groupId>


planner/core/LICENSE
```diff
@@ -0,0 +1,202 @@
...
+      License. However, in accepting such obligations, You may act only
+      on Your own behalf and on Your sole responsibility, not on behalf
+      of any other Contributor, and only if You agree to indemnify,
+      defend, and hold each Contributor harmless for any liability
+      incurred by, or claims asserted against, such Contributor by reason
+      of your accepting any such warranty or additional liability.
+
+   END OF TERMS AND CONDITIONS
+
+   APPENDIX: How to apply the Apache License to your work.
```
adriannieto: Please remove this license file, as it's already included in the root of the project (also it's not initialized properly), in my opinion it's enough with the main License file and including the header in every required file. @andreaturli  could you help us to find how this must be done?.

adriannieto:
Some considerations global considerations (I cannot comment in some files):
- Why are you using a Makefile? (planner/core/Makefile) 
- Why do you have a outdated code folder? (planner/core/old/)

.gitmodules
```diff
@@ -0,0 +1,3 @@
+[submodule "tosca-yaml-parser"]
+	path = planner/core/tosca-yaml-parser
+	url = git@github.com:pqnet/tosca-yaml-parser.git
```
mbarrientos: This is breaking the checkout of this branch. Do we need to use gitmodules? What about including that project as a Maven dependency?

planner/core/src/main/java/seaclouds/planner/Optimizer.java
```diff
@@ -0,0 +1,236 @@
...
+import com.google.common.collect.Sets;
+import seaclouds.utils.toscamodel.*;
+import seaclouds.utils.toscamodel.basictypes.IValueInteger;
+
+import java.util.*;
+
+/**
+ * Created by pq on 26/04/2015.
+ */
+public class Optimizer {
```
mbarrientos: Is this class different from the existing Optimizer (https://github.com/SeaCloudsEU/SeaCloudsPlatform/blob/master/planner/optimizer/src/main/java/eu/seaclouds/platform/planner/optimizer/Optimizer.java) ?

szenzaro:
Right comments. Thank you. I will close this pull request and make an updated version that takes into account your suggestions.

adriannieto:
@szenzaro This is not necessary, if you push new commits to the branch where the PR is done, the PR is updated as well. Could you close the new one and reopen this PR, so that the conversation is not lost?

szenzaro:
ok, I did not know that

planner/core/pom.xml
```diff
@@ -1,48 +1,105 @@
...
+                    <serverId>ossrh</serverId>
+                    <nexusUrl>https://oss.sonatype.org/</nexusUrl>
+                    <autoReleaseAfterClose>true</autoReleaseAfterClose>
+                </configuration>
+            </plugin>
+            <plugin>
+                <groupId>eu.seaclouds-project</groupId>
+                <artifactId>toscamodel</artifactId>
+                <version>1.0-SNAPSHOT</version>
+            </plugin>
```
adriannieto: Why did you use eu.seaclouds-project.toscamodel as a build plugin instead of a dependency?

planner/matchmaker/pom.xml
```diff
@@ -1,49 +1,106 @@
...
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
   <modelVersion>4.0.0</modelVersion>
-  <parent>
-    <groupId>eu.seaclouds-project</groupId>
-    <artifactId>planner</artifactId>
-    <version>0.1.0-SNAPSHOT</version>
-  </parent>
```
adriannieto: I found that you removed <parent></parent> tags, is there any particular reason (Travis CI uses this relationships in order to check if the PR passes the tests or not)

If you want to define dependencies across Maven modules, please use dependencies and parent-children hierarchy as you can see in https://github.com/SeaCloudsEU/SeaCloudsPlatform/blob/master/dashboard/pom.xml#L22

planner/core/pom.xml
szenzaro: It is already a dependency. The plugin must be deleted. You are right. ;)

planner/matchmaker/pom.xml
szenzaro: I am sorry, but I do not understand what I should change. Can you elaborate more?

adriannieto:
Travis CI cannot build because there are missing license headers. If some file doesn't need it just mark it as excluded in the pom.xml

You can check the file rat.txt in the target folder after performing a mvn clean install which contains the status of the license verification for all the files

```
[ERROR] Failed to execute goal org.apache.rat:apache-rat-plugin:0.11:check (default) on project platform: Too many files with unapproved license: 1 See RAT report in: /home/travis/build/SeaCloudsEU/SeaCloudsPlatform/target/rat.txt -> [Help 1]
````

File: planner/matchmaker/pom.xml (hunk 5 above)
Comment (adriannieto) [inline planner/matchmaker/pom.xml]: For example if you remove the parent it could lead into dependency version mismatch. Most of the shared dependencies (logger, gson...) are already included in the parent. This allows to declare a dependency for your project without the <version></version> tag. Also the parent, children dependencies allows maven to build dependent projects if necessary. So basically it helps with the dependency management.

File: planner/core/src/java/seaclouds/planner/WebServiceLayer.java [hunk 6]
```diff
@@ -0,0 +1,188 @@
...
+
+/* json parser */
+import org.json.simple.parser.*;
+import org.json.simple.*;
+
+/* Map */
+import java.util.*;
+
+
+public class WebServiceLayer extends HttpServlet {
```
adriannieto: Maybe not for this release, but take into account that using Servlets it's a very outdated way to build a REST API. Its way easier to use Jersey + Jackson as we use in https://github.com/adriannieto/SeaCloudsPlatform/blob/feature/dropwizard/dashboard/src/main/java/eu/seaclouds/platform/dashboard/resources/DeployerResource.java which eases a lot the handling of the parameters and repsonses.

mbarrientos:
It seems that there are conflicts with the current master. This is probably caused because the `planner` branch got out-dated from the recent master changes. You probably need to rebase your branch on top of the master to solve this: (https://github.com/edx/edx-platform/wiki/How-to-Rebase-a-Pull-Request)

planner/core/pom.xml
```diff
@@ -17,32 +16,106 @@ See the License for the specific language governing permissions and
...
-      <version>0.1.0-SNAPSHOT</version>
-    </dependency>
-    <dependency>
-      <groupId>eu.seaclouds-project</groupId>
-      <artifactId>planner-utils</artifactId>
-      <version>0.1.0-SNAPSHOT</version>
-    </dependency>
-  </dependencies>
-</project>
+    <artifactId>core</artifactId>
```
mbarrientos: `artifactId` should stay like planner-core, otherwise this module would become "eu.seaclouds-project.core"

planner/core/pom.xml
```diff
@@ -17,32 +16,106 @@ See the License for the specific language governing permissions and
...
+    <version>1.0-SNAPSHOT</version>
+    <parent>
+        <groupId>eu.seaclouds-project</groupId>
+        <artifactId>planner</artifactId>
+        <version>1.0-SNAPSHOT</version>
```
mbarrientos: About the version, I think that we should keep the same one along all the modules of the project. Since we don't have a complete guideline for versioning I don't know which pattern should we follow (`0.1.0-SNAPSHOT` vs `1.0-SNAPSHOT`). 

@andreaturli any ideas?

andreaturli: It is obviously up to developers of the component: if they feel like the `planner` implementation is stable and deserves a `1.0` version, go for it!

Also reading a bit http://semver.org/ may help this decision. Hope this helps

szenzaro: 0.2 would probably be a better choice

planner/core/pom.xml
szenzaro: I agree

planner/core/src/java/seaclouds/planner/WebServiceLayer.java
szenzaro: I agree

.travis.yml
```diff
@@ -1,4 +1,6 @@
 language: java
+jdk:
+  - oraclejdk8
```
andreaturli: any reason to build the entire `SeacloudsPlatform` project using jdk8?

dinitto:
Ciao Simone, the usage of jdk8 is impossible at this stage as the continuous integration chain is configured for jdk7 and all other components use jdk7. Basically, if you keep the jdk8 constraint, your component, even in the case it gets in the main repository, cannot be built and packaged with the automatic mechanisms. I'm not sure if this blocks also the packaging of the other components or not, but it certainly creates issues to the seamless creation and update of the whole platform that includes your component. Do you have in mind any solution to this problem?

.travis.yml
adriannieto: It was my suggestion, if you don't raise the Travis version the planner cannot be built with java 8 tag in maven
El 15/5/2015 4:35 p. m., Andrea Turli <notifications@github.com> escribió:In .travis.yml:
> @@ -1,4 +1,6 @@
>  language: java
> +jdk:
> +  - oraclejdk8

any reason to build the entire SeacloudsPlatform project using jdk8?

planner/core/pom.xml
adriannieto: Please could you fix it this?

PR #44, opened 2015-05-13
Fixed unignored files and libraries

mbarrientos:
Some test generated files in planner were not included in .gitignore, as well as the manually downloaded libraries from monitor.


planner/optimizer/.gitignore
```diff
@@ -4,5 +4,4 @@ target/
 .classpath
 .project
 .metadata
-.settings/
-src/test/java/eu/seaclouds/platform/planner/optimizerTest/resources/output*
```
perezp: if this line "src/test/java/eu/seaclouds/platform/planner/optimizerTest/resources/output*" is removed, but the output files of the test in OptimizerTest.java are still generated in 

private static final String OUTPUT_FILENAME       = "./src/test/java/eu/seaclouds/platform/planner/optimizerTest/resources/target/output";

I think that the output files will continue being included in the commits.

mbarrientos: They stay ignored because everything under target/ is ignored too. It looked like the folder "resources" contained 2 different kind of files, input files and generated ones.

perezp: If you say so, OK for me. 

I'm not so sure yet, because target/ is not included in that .gitignore (it had been included in the .gitignore of the monitor, not the one of the optimizer) . The highest level .gitignore of the platform contains the target/ , but this is not the same folder as  ./src/test/java/eu/seaclouds/platform/planner/optimizerTest/resources/target/ . 

In any way, even if something goes wrong, it is less costly in time to accept the PR, and I will change the path where the the output files are written to a place "gitignored" in my next PR…. than to discuss this in the comments.

perezp: +1

perezp:
+1

planner/optimizer/.gitignore
mbarrientos: @perezp There's no need to blind-trust me!. If you want to check my PR properly you should checkout my branch and try it yourself.

However, target is already included in your gitignore (https://github.com/SeaCloudsEU/SeaCloudsPlatform/blob/master/planner/optimizer/.gitignore#L3). Also, you don't need to specify the full path to ignore a file: http://stackoverflow.com/questions/1470572/gitignore-ignore-any-bin-directory.

However, thanks for reviewing it!

perezp: Thank you for the lesson about gitignore!

PR #45, opened 2015-05-15
Monitor

dionysiosAthanasopoulos:


adriannieto:
In my opinon if the monitor doesn't need a landing page, ```monitor/WebContent/index.jsp``` should't be necesary

monitor/src/main/java/core/UnixFileExecution.java
```diff
@@ -0,0 +1,62 @@
...
+			
+
+			p.waitFor();
+		}
+
+		catch (Exception ex) {
+
+			ex.printStackTrace();
+
+			System.exit(-1);
```
adriannieto: What happens if an exception happens? The whole monitor finishes his execution? This is a very dangerous behavior in my opinion.

dinitto:
We decided not to use this API in the SeaClouds platform

PR #46, opened 2015-05-20
Feature/common

andreaturli:
- general clean up to the maven structure: commented monitor module, among other fixes, as not used
- add `common` module where the seaclouds app and the policies to install data collectors are defined
- I will move `brooklyn-modaclouds` entities to `common` there as soon as this gets merged

@MicheleGuerriero could you please take a look, particularly to the mechanism that installs the data collector for sigar?

### DIFF
--- a/common/src/test/resources/mysql-catalog.yaml
+++ b/common/src/test/resources/mysql-catalog.yaml
@@ -1,0 +1,18 @@
+brooklyn.catalog:
+  id: eu.seaclouds-project.mysql
+  version: 1.0
+  description: A Tomcat app-server
+  displayName: Tomcat
+  iconUrl: classpath:///tomcat-logo.png
+  itemType: template
+
+  item:
+    services:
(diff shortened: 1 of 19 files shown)


common/src/test/resources/mysql-catalog.yaml
```diff
@@ -0,0 +1,18 @@
...
+  id: eu.seaclouds-project.mysql
+  version: 1.0
+  description: A Tomcat app-server
+  displayName: Tomcat
+  iconUrl: classpath:///tomcat-logo.png
+  itemType: template
+
+  item:
+    services:
+    - type: io.cloudsoft.amp.canopy.compose.apps.CanopyComposeApp
```
mbarrientos: I guess this should be a SeaCloudsApp as well, right? @andreaturli

mbarrientos:
Even though I couldn't run the live tests completely it looks good to me.

andreaturli:
thanks @mbarrientos

PR #47, opened 2015-06-01
FEATURE (Dashboard): Updates on dashboard backend 

adriannieto:
This updates the existing service layer of the dashboard with the following changes:
* We refactored the whole project to use `dropwizard.io`.
* Dashboard is now communicating only through REST calls to the rest of SeaClouds components.
* Minor changes on the frontend (`deployer.html`) to accept loading of `monitoring rules` and `sla agreements`. 

Also, this PR should fix master build problems due to old maven dependency between `planner` and `dashboard`.


dashboard/src/main/java/eu/seaclouds/platform/dashboard/resources/MonitorResource.java
```diff
@@ -0,0 +1,205 @@
...
+import java.net.URISyntaxException;
+import java.util.Iterator;
+
+@Path("/monitor")
+@Produces(MediaType.APPLICATION_JSON)
+public class MonitorResource {
+
+    @GET
+    @Path("metrics/value")
+    public Response getMetric(@QueryParam("applicationId") String applicationId,
```
andreaturli: is it better to have a custom Response or directly the monitorResponse object?

adriannieto: As the seaclouds upper layer it's not defined I think it's better to just proxy raw responses until the model will be well defined. Then we will let Jackson + Gson to translate Objects into Responses.

andreaturli:
hi @adriannieto very great clean up, before a proper review I've a couple of general comments:

* there are heterogenous commits in this PR (downgrade java, probably deserves its own PR): we should avoid this 
* You implemented a rest http client for brooklyn as far as I can see: why haven't you used the [brooklyn-rest-client](https://repository.apache.org/content/repositories/snapshots/org/apache/brooklyn/brooklyn-rest-client/) ?
* I've seen you often apply the following pattern in the rest client: `public Response getX(...)` do you think it is better than having something like `public X getX(...)`

adriannieto:
* If I open a PR just for the Java version the master branch will not build probably. 
* brooklyn-rest-client generates a collision between dependencies (as it's probably autogenerated by Swagger v1, it uses legacy libraries that makes the dashboard to fail on runtime)
* This one I think it's already answered in the previous comment (I didn't read this before answering)

Thank you very much @andreaturli

andreaturli:
Thanks @adriannieto, could you share a gist that shows the dep problem when using the rest client?
Also, do you think you can help in fixing the problem with `brooklyn-rest-client` as I think it is better to have a solid official client than trying to implement one ad-hoc per seaclouds? Of course, not in this PR, as it deserves a dedicated discussion, I think.

adriannieto:
@andreaturli it was @mbarrientos who realized about this problem maybe he can help us with the problem. The thing is the sooner or later if we use seaclouds upper layer a translation must be done in some point into our own namespace, so I don't know it the usage objects instead bare JSON will ease or not the the translation process.

adriannieto:
@andreaturli about ``maven-compiler-plugin`` it belongs to planner submodule, should we modify it on this PR (although we already did something similar with the Java version)

One thought about downgrading Java, if some of the projects declared as dependencies requires Java 8 it will build but fail on runtime. 

What do you mean about moving ConfigParameters to a yaml file? How is going to be deployed with seaclouds-distribution then?

andreaturli:
@adriannieto for example, I'm thinking about using a configuration file like [yaml](https://github.com/apache/incubator-brooklyn/blob/master/software/database/src/main/resources/brooklyn/entity/database/crate/crate.yaml) which contains placeholders that are substitued [here](https://github.com/apache/incubator-brooklyn/blob/master/software/database/src/main/java/brooklyn/entity/database/crate/CrateNodeSshDriver.java#L69)

andreaturli:
Thanks @adriannieto I've raised a couple of minor comments, but I think we can address them in a subsequent PR.
+1

merging now, thanks

PR #50, opened 2015-06-02
Maven dependencies cleanup

mbarrientos:
* Pulling up to parent most of the version numbers.
* Removed global dependencies and added `dependencyManagement` on parent.


adriannieto:
@mbarrientos  Is this still necessary?

PR #51, opened 2015-06-02
fix javadoc generation even with jdk8

andreaturli:
- it adds a profile that is active only when java8 is used to build the project to avoid failures during the javadoc generation as per http://stackoverflow.com/questions/15886209/maven-is-not-working-in-java-8-when-javadoc-tags-are-incomplete

@mbarrientos could you please take a look?


andreaturli:
@adriannieto could you please have a look at this build https://travis-ci.org/SeaCloudsEU/SeaCloudsPlatform/builds/65145827 ?
looks like the failure of this #51 is due to a dashboard test, thanks

andreaturli:
@adriannieto `HttpRequestBuilderTest.testGetRequestWithEcho` is using an external service http://echo.jsontest.com/ which shows now 503 Over Quota 

I think you may consider https://github.com/square/okhttp/tree/master/mockwebserver for these kind of tests.

andreaturli:
close and re-opening to trigger travis CI

mbarrientos:
Looks fine. I would remove the coveralls badge from the readme as well (at least for now...). +1

adriannieto:
@andreaturli Why it's building right now? Did you changed this? Should I open a new PR to fixit?

andreaturli:
@adriannieto the external service has a quota per day, this morning was not yet over quota.

A fix would be appreciated, but not for this PR

mbarrientos:
Thanks @andreaturli. Merging.

PR #53, opened 2015-06-03
FEATURE:First SeaClouds Object Model Interface

szenzaro:
@adriannieto can you look at this?

### DIFF
--- a/common/src/main/java/eu/seaclouds/common/objectmodel/AppModule.java
+++ b/common/src/main/java/eu/seaclouds/common/objectmodel/AppModule.java
@@ -1,0 +1,109 @@
+/**
+ * Copyright 2014 SeaClouds
+ * Contact: SeaClouds
+ *
+ *    Licensed under the Apache License, Version 2.0 (the "License");
+ *    you may not use this file except in compliance with the License.
+ *    You may obtain a copy of the License at
+ *
+ *        http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *    Unless required by applicable law or agreed to in writing, software
+ *    distributed under the License is distributed on an "AS IS" BASIS,
+ *    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *    See the License for the specific language governing permissions and
+ *    limitations under the License.
+ */
+package eu.seaclouds.common.objectmodel;
+
+import eu.seaclouds.common.objectmodel.apptypes.AppType;
+import eu.seaclouds.common.objectmodel.constraints.Constraint;
+import eu.seaclouds.common.objectmodel.hosttypes.HostType;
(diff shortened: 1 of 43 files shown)


adriannieto:
Thank you very much for reopening the PR. As this PR is very important I would like to know the opinion of the other partners involved, regarding to the model that you defined based on the Pisa meeting results.

The code looks good, and you already provided tests. However due this PR it's a starting point of the SeaClouds Object Model there are some empty implementations and I don't know how to handle this. In my opinion I would merge it and then we will continue the implementation with new PR although everything in master branch must be "ready to deploy code", so I will forward this issue to @andreaturli  who is more expertise in good-practices on Github.

I will hold my +1 until other partners share their opinion.

Thank you very much!

szenzaro:
The idea was to get a first impression on the interface. There are still open modeling issues for the interface i guess. 
The implementation is left empty on purpose since it will probably be a shared task.

Finally the tests are still not complete and are meant to provide some first look at how we will use the interface. We will need further test in my opinion.

adriannieto:
@perezp @rosogon Could you share your opinion?

marcoriol:
Thank you all of you guys :)

It would be great if you could check/contribute on the elements related to the optimizer (for @perezp) and QoB (for @rosogon). You will find in common/src/test/java/eu/seaclouds/common/compose/objectmodel/ObjectModelTests.java some tests that are useful as example of usage of this model. Particularly, you will see examples of usage of this model for the Optimizer and for the generation of the AAM according to what was defined in the requirements document for the upper layer, but for some methods we need from you some feedback/clarifications. Nevertheless, as this model is still under development, don't look at it as written in stone. Contributions are obviosuly more than welcome :). Also take a look at the elements of the model, and provide us some feedback on how it should be defined: the QoS concepts used by the Optimizer(for @perezp) and QoB/SLA concepts (for @rosogon). 

Many thanks :). If you have any doubt don't hesitate to ask :)

perezp:
I mainly agree with @adriannieto message. For me it looks good. 

We cannot completely know how good it will work, but we should merge it to start working with this object model to be able to gain experience, fill the empty code and/or propose further refinements. 

Github says that it cannot be automatically merged. Once this issue is solved, I would have no problem to give the +1 if everything continues clean, even if it is an initial version with empty code.

adriannieto:
Please rebase the PR  in order to fix this problems (also please join the two commits)

szenzaro:
The conflict is on the .git ignore file alone. Am I allowed to merge to the master after resolving this conflict and squashing the two commits?

adriannieto:
As we need to start working on this, you can consider that it's already has +1 @rosogon  will be able to open issues/pr to improve this work later.

szenzaro:
Now it should be fine. Is it is ok, feel free to merge.

PR #54, opened 2015-06-03
Planner: refactoring + packaging (WAR)

buccarel:
I refactored the code of the Planner in such a way to be compliant (in directories and packaging) together with other SeaClouds components.
The generation of the .WAR package file is also automatic and working.


andreaturli:
@buccarel thanks for this PR

A couple of preliminary comments: 
- the PR can't be automatically merged, can you investigate why?
- I can see you removed almost 50 files from the planner without replacing them: was it intentional?

buccarel:
Fixing very small issues.

PR #55, opened 2015-06-03
Fix/planner

buccarel:
After fixing small issues, pull request that automatically merges.
- refactored code
- enabled the automatic generation of the .war artifact


perezp:
Thank you for the new PR. 
It looks like the issue pointed out in the previous PR regarding the removed files still happens in this new one.

buccarel:
@perezp After my unintended mistake, I restored your optimizer as it was before. Since the your optimizer has been involved, would you review this PR, please? The reason why I did this is because the code of the Planner needed to be refactored in terms of directories in order to let the war maven plugin work correctly.

perezp:
@buccarel I'll review it. It's not yet clear to me the issue about java version, and I cannot compile this PR now. 

Nevertheless, this PR is not changing any version. Version inconsistencies are already in the master branch. 

Are you happy with changing the version of the java in the pom.xml in the Planner? Or just remove it and it will use the version of its parent. (I know that in this PR you are not changing the version in the pom.xml of the planner, but in this PR the previous inconsistency becomes "active")

buccarel:
@perezp personally, I don't really have any issues about the version of the used jdk. If it depended on me, you could've put any version you liked more. I don't know what kind of issue there currently is about it.

About the compilation, I don't know why you can't compile it: Travis did not show any error on last commit and I am also able to compile it on both my laptop and Simone's.

When talking about review, I was more focusing on the directory structure or the code itself. I was not aware of any other problem. Maybe @andreaturli can give some hint about the problems that you're experiencing?

perezp:
This is the error I get with "mvn clean install":

[ERROR] COMPILATION ERROR : 
[INFO] -------------------------------------------------------------
[ERROR] Failure executing javac, but could not parse the error:
javac: invalid target release: 1.8

szenzaro:
@perezp can you try to edit the planner pom.xml from:

<groupId>org.apache.maven.plugins</groupId>
                <artifactId>maven-compiler-plugin</artifactId>
                <version>2.3.2</version>
                <configuration>
                    <source>1.8</source>
                    <target>1.8</target>
                </configuration>

to 

<groupId>org.apache.maven.plugins</groupId>
                <artifactId>maven-compiler-plugin</artifactId>
                <version>2.3.2</version>
                <configuration>
                    <source>1.7</source>
                    <target>1.7</target>
                </configuration>
and see if it fixes the issue?

perezp:
@szenzaro , yes, it does. I suggested such change in the first comment. 

As also suggested, another way to fix the issue is to remove the maven-compiler-plugin from the plugin list, since it will use the declaration given in the parent pom

szenzaro:
@perezp perfect! Other issues?

buccarel:
@perezp I removed the plugin node in the pom.xml as you suggested and everything is working fine. Thank you.

perezp:
Now it compiles fine with java7. Thanks! 

I've reviewed the refactoring. 
The following image shows the folder structure I see for the planner that is proposed by this PR 
 https://drive.google.com/file/d/0By0YQQwm91rdSTlJMjBXeEYxeWs/view?usp=sharing 

Was your intention to create this concrete structure? 
Can you explain a bit the objective of this refactor? Or is there still some error? 

I see some parts misplaced but maybe it is due to an error that happened by chance.

buccarel:
@perezp well, there are 2 perspectives to consider:
1) most important - I refactored the structure of the directories of the planner in such a way that the war-plugin is able to create the final package consistently. It indeed has some requirements (like the existance of the src/main/java hierarchy);
2) I refactored the location of the Planner mainly to be compliant with the other modules that are already present in the SeaCloudsEU repo. In particular, in reference to your optimizer (just because it's involved together with the planner), the directory "optimizer" was under "/planner/", that is outside of "src", since it has its own "src" directory within.

So, in conclusion: yes, the structure of the planner (not the optimizer) that you're seeing right now is exactly intended for the 2 reasons above. So, unless I made some other mistake, that should be the concrete and final shape.

ADDENDUM.
Yesterday I wrote to your skype to inform you about this and I also asked you if you can kindly refactor the optimizer structure to be compliant to it too (task to do after my PR is accepted, I guess). Otherwise, the war-plugin will not include the optimizer in the final package.

Thank you very much.

PR #57, opened 2015-06-05
prepare release

andreaturli:
- add instructions on how to release the code

@kiuby88 could you please review just to see if the instructions are clear enough?

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,6 +19,20 @@ Getting Started
 * Installation guide: TODO
 * Quickstart: TODO
 
+How to release it!
+-------------------
+In order to release a new version:
+
+1. create a new `release` branch out of `master` branch
+2. update the maven version inside `release/new-branch` pom.xml:
+
+    $ find . -iname pom.xml | xargs sed -i '' "s;<version>0.1.0-SNAPSHOT</version>;<version>0.1.0-M19</version>;g"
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -19,6 +19,20 @@ Getting Started
...
+In order to release a new version:
+
+1. create a new `release` branch out of `master` branch
+2. update the maven version inside `release/new-branch` pom.xml:
+
+    $ find . -iname pom.xml | xargs sed -i '' "s;<version>0.1.0-SNAPSHOT</version>;<version>0.1.0-M19</version>;g"
+
+3. mvn clean install
+
+If everything is ok, finally push it to Sonatype OSSRH:
```
kiuby88: I think it should be the fourth step and the mvn command should be indented (as second step and `$ find . -iname pom.xml | xargs sed -i '' "s;<version>0.1.0-SNAPSHOT</version>;<version>0.1.0-M19</version>;g"`)

rosogon:
As an alternative to sed, we can use http://www.mojohaus.org/versions-maven-plugin/

mvn versions:set -DnewVersion=0.1.0-M19 && mvn versions:commit

andreaturli:
@rosogon thanks for your suggestion

PR #58, opened 2015-06-05
FIX(Dashboard): Notify the SLA after the rules are installed.

### LINKED ISSUE #56: The SLA is not notified about the rules installation
Testing the SeaClouds platform and trying to show also the SLA working it seems theSLA is not notified after the rules installation and we still need to notify the SLA manually. Also looking at the code here  https://github.com/SeaCloudsEU/SeaCloudsPlatform/blob/master/dashboard/src/main/java/eu/seaclouds/platform/dashboard/resources/DeployerResource.java#L67 it looks like the Dashboard send the agreements but does not notify the SLA about the rules installation. In order to perform this notification the following REST call needs to be performed: POST SLA-IP:8080/seaclouds/commands/rulesready. @adriannieto

adriannieto:
This Fixes #56.


andreaturli:
+1

Issue #48, opened 2015-06-01
Replace the ENV vars of the Dashboard with a YAML file


adriannieto:
Regarding to the review of  https://github.com/SeaCloudsEU/SeaCloudsPlatform/pull/47 we found an interesting way to replace enviroment variables that wires the entire project with the Dashboard.  Maybe as we are using Dropwizard.io to deploy the dashboard, we can use the configuration YAML to setup all the necessary parameters.

adriannieto:
Closed by #60

PR #60, opened 2015-06-05
Added support for Dropwizard configuration on dashboard

mbarrientos:
Added support for Dropwizard configuration on dashboard:
- config.yml is now bootstrapped with environment variables.
- REST resources access the endpoint through factories providing config information.


mbarrientos:
This fixes #48.

dashboard/src/main/java/eu/seaclouds/platform/dashboard/DashboardApplication.java
```diff
@@ -34,16 +36,25 @@ public static void main(String[] args) throws Exception {
 
     @Override
     public void initialize(Bootstrap<DashboardConfiguration> bootstrap) {
+        // Setting configuration from env variables
+        bootstrap.setConfigurationSourceProvider(
+                new SubstitutingSourceProvider(bootstrap.getConfigurationSourceProvider(),
+                        new EnvironmentVariableSubstitutor(false)
```
andreaturli: why `strict=false` ?

dashboard/config.yml
```diff
@@ -4,4 +4,15 @@ server:
     port: 8000
   adminConnectors:
   - type: http
-    port: 8001
\ No newline at end of file
+    port: 8001
+deployer:
+  host: ${DEPLOYER_HOST}
```
andreaturli: In this way we are still substituting those values using ENV VAR, which is fine but not ideal.
When I suggested to use Dropwizard configuration I was expecting to have a `config.yml.template` where values were replaced by Brooklyn when known, but we'd need a dashboard entity for that, so let's keep it as is.

Probably the idiomatic way in Dropwizard is to define your own `ConfigurationSourceProvider` but not sure.

mbarrientos: The only problem to using Brooklyn yaml templating is that it makes the Dashboard a bit more Brooklyn dependent (which is not a big problem!). Maybe we could support two templates or something similar.

However, this template is OK for building with any kind of distribution ATM. I'll add this feature when the brooklyn dashboard entity is ready, which will be soon.

dashboard/src/main/java/eu/seaclouds/platform/dashboard/DashboardApplication.java
mbarrientos: I let it false to support empty fields on the user/pass parameters.

mbarrientos:
@andreaturli I opened https://github.com/SeaCloudsEU/SeaCloudsPlatform/pull/62 to add brooklyn config support to the actual seaclouds distribution (dashboard only). 
Right now, the dashboard is able to build from env vars when launching as a standalone and with brooklyn config keys when using Brooklyn Entity.

If you consider it ok we can merge this and move forward with https://github.com/SeaCloudsEU/SeaCloudsPlatform/pull/62

andreaturli:
@mbarrientos I think it is really good, let's merge that.

I've a quick look at #62, not sure `usage/installer` is the right place for the new entities (also modaclouds entities should be moved, my bad) but the implementation looks exactly what we need

PR #61, opened 2015-06-08
Add alien4cloud tosca parser/serializer

andreaturli:


common/src/test/java/eu/seaclouds/common/apps/AbstractSeaCloudsAppTest.java
```diff
@@ -14,7 +14,7 @@
  *    See the License for the specific language governing permissions and
```
kiuby88: I am not sure about if this test and the subclasses make sense in this PR, probably a new PR should be open.

andreaturli: ok, it makes sense.

common/src/test/java/eu/seaclouds/common/tosca/ToscaParserSupplierTest.java
```diff
@@ -0,0 +1,57 @@
...
+    @Test
+    public void testNodeType() throws URISyntaxException, ParsingException {
+        ToscaParser toscaParser = new ToscaParserSupplier().get();
+        ParsingResult<ArchiveRoot> parsingResult = toscaParser.parseFile(Paths.get(Resources.getResource("tosca/SimpleProfile_wd03/parsing/tosca-node-type.yml").toURI()));
+        ArchiveRoot archiveRoot = parsingResult.getResult();
+        assertNotNull(archiveRoot);
+        Assert.assertNotNull(archiveRoot.getArchive());
+        Assert.assertEquals(TOSCA_VERSION, archiveRoot.getArchive().getToscaDefinitionsVersion());
+        Assert.assertEquals(1, archiveRoot.getNodeTypes().size());
+    }
```
kiuby88: Probably, we should to add more test in order to check the applicationParting. For example:

In the NodeType, the relationshipType of the requirements retrieves null, so it is not get the correct value from the YAML, probably because the used type does not exist in the yaml example.

common/pom.xml
```diff
@@ -19,97 +19,151 @@
...
+    <description>SeaClouds common module</description>
+
+    <properties>
+        <spring.version>4.1.4.RELEASE</spring.version>
+        <commons-lang.version>3.3.2</commons-lang.version>
+        <validation-api.version>1.1.0.Final</validation-api.version>
+    </properties>
+
+    <repositories>
+        <repository>
```
kiuby88: I think it may be a good idea if the issue about mvn version is commented on the Readme. Indicating that to build this project is required at least mvn 3.2.5

common/src/test/java/eu/seaclouds/common/tosca/ToscaParserSupplierTest.java
andreaturli: not sure I get the comment about `relationshipType`

common/pom.xml
andreaturli: added that to the README

kiuby88:
+1

PR #62, opened 2015-06-10
FEATURE(Installer): added Brooklyn entity for dashboard.

mbarrientos:
This relies on the configuration proposed at https://github.com/SeaCloudsEU/SeaCloudsPlatform/pull/60. 

When building with a SeaClouds YAML blueprint the config file is populated from Brooklyn `ConfigKeys`.


andreaturli:
@mbarrientos maybe we need to move `seaclouds-entities` to a dedicated maven module rather than in the installer. wdyt?

mbarrientos:
@andreaturli I agree. Do you mean another child module to `usage` right? What about `SeaCloudsApp`, which is under `common` package? should we move it as well?

andreaturli:
It is not really `usage` IMHO. I think `common` was intended for things that are useful to more than one maven module, so maybe under `common` we could add another package called `entities` or re-structure `common` to have children like `objectmodel`, `tosca`, `entities`, `apps`, `policies` but maybe it is over-engineering.

Let's start by adding a new package to `common` called `entities` where moving `modaclouds` and `dashboard` entities, but we can keep in mind that we may want to refactor it at some point. Makes sense?

mbarrientos:
Moving it into `common` was my first thought too, so let's work with a single maven module and keep it simple ATM. We will be refactoring whenever we need to extract one single dependency from it in the future. 
I'm opening a new PR for moving all the entities at once, is that ok?

andreaturli:
+1

usage/installer/src/main/java/eu/seaclouds/dashboard/SeacloudsDashboardSshDriver.java
```diff
@@ -0,0 +1,125 @@
...
+import com.google.common.collect.ImmutableList;
+import com.google.common.collect.ImmutableMap;
+import java.net.URI;
+import java.util.List;
+
+import static java.lang.String.format;
+
+public class SeacloudsDashboardSshDriver extends JavaSoftwareProcessSshDriver implements SeacloudsDashboardDriver {
+
+    private static final String CONFIG_FILE_NAME = "config.yml";
```
andreaturli: would it make sense to make this a ConfigKey as well, maybe with a default value?

andreaturli:
@mbarrientos can you squash the 2 commits into 1, rebase and push it again?

andreaturli:
lgtm +1

Merging now

Thanks @mbarrientos

PR #64, opened 2015-06-11
fix planner maven structure

andreaturli:
uncomment optimizer
enable build for matchmaker and discoverer
fix maven layout for matchmaker,optimizer and discoverer
fix compilation errors (wtf!?)
fix parents in multiple pom.xml
remove unnecessary build plugins as inherited from parents
remove jdk8 requirement

### DIFF
--- a/planner/core/pom.xml
+++ b/planner/core/pom.xml
@@ -20,28 +20,26 @@ limitations under the License.
          xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
     <modelVersion>4.0.0</modelVersion>
     <parent>
-    <groupId>eu.seaclouds-project</groupId>
-    <artifactId>planner</artifactId>
-    <version>0.1.0-SNAPSHOT</version>
-  </parent>
-  <artifactId>planner-core</artifactId>
-  <version>0.1.0-SNAPSHOT</version>
-  <name>planner-core</name>
-  <url>http://maven.apache.org</url>
-  <properties>
-    <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
-  </properties>
-  <dependencies>
-    <dependency>
-      <groupId>eu.seaclouds-project</groupId>
-      <artifactId>matchmaker</artifactId>
-      <version>0.1.0-SNAPSHOT</version>
-    </dependency>
-    <dependency>
-      <groupId>eu.seaclouds-project</groupId>
-      <artifactId>planner-utils</artifactId>
-      <version>0.1.0-SNAPSHOT</version>
-    </dependency>
-  </dependencies>
+        <groupId>eu.seaclouds-project</groupId>
+        <artifactId>planner</artifactId>
+        <version>0.1.0-SNAPSHOT</version>
+        <relativePath>../pom.xml</relativePath>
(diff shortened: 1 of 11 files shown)


andreaturli:
Oh dear, what a mess!
This PR tries to fix what it looks more like a train wreck to me!

TODO
As the tosca-parser from a4c we need to remove the dependency on the old tosca-parser which is even coming from an external repository

perezp:
Thank you for the PR!

Something new is happening in the tests of modules inside the planner, concretely in the optimizer. I do not see their outputs, although the report says that none of the tests failed. It happens in both my local machine and travis. It say this: 

-------------------------------------------------------
 T E S T S
-------------------------------------------------------
Running TestSuite
SLF4J: Failed to load class "org.slf4j.impl.StaticLoggerBinder".
SLF4J: Defaulting to no-operation (NOP) logger implementation
SLF4J: See http://www.slf4j.org/codes.html#StaticLoggerBinder for further details.
Tests run: 7, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 2.998 sec

Results :

Tests run: 7, Failures: 0, Errors: 0, Skipped: 0

================================================================

I looked around for the solution. I found that adding the following dependency in the pom.xml , tests are shown again. 

<dependency>
                <groupId>org.slf4j</groupId>
                <artifactId>slf4j-simple</artifactId>
                <version>1.7.12</version>
</dependency>


Maybe it's not the proper solution. If it's a problem of the tests, let me know (the last time I checked the outputs from travis, tests were correctly shown without adding the previous dependency… I don't know). 
Please, @andreaturli let me also know if this is not even a matter of this PR but it's something that has to be dealt with in the next PR of the optimizer.

perezp:
Thank you for the PR!

Something new is happening in the tests of modules inside the planner, concretely in the optimizer. I do not see their outputs, although the report says that none of the tests failed. It happens in both my local machine and travis. It say this:

T E S T S

Running TestSuite
SLF4J: Failed to load class "org.slf4j.impl.StaticLoggerBinder".
SLF4J: Defaulting to no-operation (NOP) logger implementation
SLF4J: See http://www.slf4j.org/codes.html#StaticLoggerBinder for further details.
Tests run: 7, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 2.998 sec

Results :

Tests run: 7, Failures: 0, Errors: 0, Skipped: 0

I looked around for the solution. I found that adding the following dependency in the pom.xml , tests are shown again.


`<dependency>`
`                <groupId>org.slf4j</groupId>`
 `               <artifactId>slf4j-simple</artifactId>`
`                <version>1.7.12</version>`
` </dependency>`

Maybe it's not the proper solution. If it's a problem of the tests, let me know (the last time I checked the outputs from travis, tests were correctly shown without adding the previous dependency… I don't know).
Please, @andreaturli let me also know if this is not even a matter of this PR but it's something that has to be dealt with in the next PR of the optimizer.

andreaturli:
@perezp are you happy with it now? can we merge it?

perezp:
@andreaturli Yes, I'm happy with it. 
 
Now all the version numbers are coherent. I had not merged it during these last days because I wanted to know if planner developers were also happy with the version number you chose  ``` 0.1.0-SNAPSHOT ```  for the planner module.

PR #65, opened 2015-06-12
Updates in usage/installer to support MODAClouds TOWER4 Clouds

MicheleGuerriero:
@perezp @andreaturli @rosogon 

MODAClouds Brooklyn entities modified in order to support the the new version of the monitoring, TOWER4 CLouds:

- modified ssh driver and the main class the data analyzer entities to retrieve release 0.2 and be able to launch and configure it with the available parameters
- modified ssh driver and the main class within the monitoring manager entities to retrieve  release 0.2 from tower4clouds repo and be able to launch and configure it with the available parameters
- converted all the knoledge base entities into the history db entities. Now it is possible to retrive release 0.2 of the history dbfrom tower4clouds repo and to launch and configure it with the available parameters

### DIFF
--- a/usage/installer/src/main/java/eu/seaclouds/modaclouds/dda/MODACloudsDeterministicDataAnalizerSshDriver.java
+++ b/usage/installer/src/main/java/eu/seaclouds/modaclouds/dda/MODACloudsDeterministicDataAnalizerSshDriver.java
@@ -48,7 +48,7 @@ public class MODACloudsDeterministicDataAnalizerSshDriver extends JavaSoftwarePr
    @Override
    public void preInstall() {
       resolver = Entities.newDownloader(this);
-      setExpandedInstallDir(Os.mergePaths(getInstallDir(), resolver.getUnpackedDirectoryName(format("rsp-services-csparql-%s-modaclouds", getVersion()))));
+      setExpandedInstallDir(Os.mergePaths(getInstallDir(), resolver.getUnpackedDirectoryName(format("data-analyzer-%s", getVersion()))));
    }
 
    @Override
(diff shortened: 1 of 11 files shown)


usage/installer/src/main/resources/brooklyn/entity/modaclouds/modaclouds.yaml
```diff
@@ -21,8 +22,5 @@ services:
     brooklyn.config:
       modaclouds.dda.ip: $brooklyn:component("dda").attributeWhenReady("host.address")
       modaclouds.dda.port: $brooklyn:component("dda").attributeWhenReady("modaclouds.dda.port")
-      modaclouds.kb.ip: $brooklyn:component("kb").attributeWhenReady("host.address")
-      modaclouds.kb.port: $brooklyn:component("kb").attributeWhenReady("modaclouds.kb.port")
-      modaclouds.kb.dataset.path: $brooklyn:component("kb").attributeWhenReady("modaclouds.kb.path")
```
mbarrientos: @MicheleGuerriero is this endpoint not needed anymore?

MicheleGuerriero: @mbarrientos The MODACLouds Knowledge Base is actually no more part of TOWER4 Clouds and it has been removed. It has been somehow replaced by the HistoryDB. Actually I did not set the link between the manager and the histoy db in the modaclouds.yaml since we are not sure about if the History DB will be used, but actually I put it in the modaclouds.yaml and so it is currently deployed. So as you suggest it requires some further configuration. Can I add this with another commit?

mbarrientos: In case that we don't use HDB then we should also remove the deployment of it. Probably we could have 2 different blueprints `modaclouds.yaml` and `modaclouds-with-hdb.yaml`.

mbarrientos: To update the PR just push more commits to your branch and GitHub will autoupdate. 
Then, once this PR is ready to merge you can squash all the commits into one, for simplicity.

MicheleGuerriero: Ok, can I simply add this with another commit on the same branch?

mbarrientos:
I think this looks good to me. Anyone from the integration who can have a look and speed up this PR? Thanks. @andreaturli @rosogon @perezp

perezp:
Nothing against merging it from my side.

javicubo:
@rosogon please check this PR and modify the rest call to connect with the SLA, and give your feedback to accept asap, as the others are happy with the merge, and we need to know if it would be accepted as we need to use this version to use Graphite to visualize the metrics of DC. Thanks.

rosogon:
It's ok for me and can be merged. sla-core is modified.

perezp:
Before merge the message from travis was "passing" but travis execution after merge has errored. Reason is "check style". 

I'll revert the PR.

PR #68, opened 2015-07-01
Solved some problems with the previous PR due to the check-style plugin 

MicheleGuerriero:
Class MODACloudsHistoryDBSshDriver had some style errors, in particular some tabs for which the maven check-style plugin was complaining. @perezp @andreaturli @rosogon


perezp:
In my opinion it can be merged. +1 
According to travis, it will solve the errors in the current master (style check).

andreaturli:
thx @MicheleGuerriero

PR #69, opened 2015-07-02
Fix seaclouds-on-byon.yaml

andreaturli:
Fix Dashboard test
Fix MODAClouds entities to implement UseJava


andreaturli:
@adriannieto could you please test that PR to check in particular if the `seaclouds-on-byon.yaml` is ok? Thanks!

adriannieto:
I found errors with the Deployer and Dashboard entity, I need to check it further.

adriannieto:
@andreaturli  this PR will be closed by #70, because I need this fixes to make the dashboard to work properly. I'm not sure if this is the right way to work when a PR requires another PR to work.

PR #72, opened 2015-07-06
FEATURE (Discoverer): Core and WS layer

buccarel:
@andreaturli @szenzaro This is the correct PR for the Discoverer. Can you please review it?
Thanks.

### DIFF
--- a/discoverer/src/main/webapp/WEB-INF/web.xml
+++ b/discoverer/src/main/webapp/WEB-INF/web.xml
@@ -1,0 +1,49 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+ Copyright 2014 SeaClouds
+ Contact: SeaClouds
+ 
+ Licensed under the Apache License, Version 2.0 (the "License");
+ you may not use this file except in compliance with the License.
+ You may obtain a copy of the License at
+ 
+ http://www.apache.org/licenses/LICENSE-2.0
+ 
+ Unless required by applicable law or agreed to in writing, software
+ distributed under the License is distributed on an "AS IS" BASIS,
+ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ See the License for the specific language governing permissions and
+ limitations under the License.
+ 
+ -->
+<web-app>
+
+  <display-name>SeaClouds Discoverer WS Layer</display-name>
+  <description>The SeaClouds's Discoverer as RESTful service</description>
+
+
+  <!-- ADD OFFERING -->
+  <servlet>
+	<servlet-name>discoverer_add_offering</servlet-name>
+	<servlet-class>eu.seaclouds.platform.discoverer.ws.AddOffer</servlet-class>
+  </servlet>
+
+  <servlet-mapping>
+	<servlet-name>discoverer_add_offering</servlet-name>
+	<url-pattern>/addoffer</url-pattern>
+  </servlet-mapping>
+  
+  
+  <!-- CRAWLER MANAGER -->
+  <servlet>
+  	<servlet-name>discoverer_crawler_manager</servlet-name>
+  	<servlet-class>eu.seaclouds.platform.discoverer.ws.CrawlerManager</servlet-class>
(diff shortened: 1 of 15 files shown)


szenzaro:
Travis doesn't compile it; It found an error at

/home/travis/build/SeaCloudsEU/SeaCloudsPlatform/discoverer/src/main/java/eu/seaclouds/platform/discoverer/core/Discoverer.java:[207,17] eu.seaclouds.platform.discoverer.core.Discoverer.OfferingRepoIterator is not abstract and does not override abstract method remove() in java.util.Iterator

szenzaro:
Now it looks ok. If @andreaturli approves it we can merge. +1

buccarel:
Since Andrea is out of the office in these days, please @mbarrientos can you review the PR in place of him?

discoverer/src/main/webapp/WEB-INF/web.xml
```diff
@@ -0,0 +1,49 @@
...
+ distributed under the License is distributed on an "AS IS" BASIS,
+ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ See the License for the specific language governing permissions and
+ limitations under the License.
+ 
+ -->
+<web-app>
+
+  <display-name>SeaClouds Discoverer WS Layer</display-name>
+  <description>trololo</description>
```
mbarrientos: ?

discoverer/src/main/webapp/WEB-INF/web.xml
```diff
@@ -0,0 +1,49 @@
...
+  <servlet>
+	<servlet-name>discoverer_add_offering</servlet-name>
+	<servlet-class>eu.seaclouds.platform.discoverer.ws.AddOffer</servlet-class>
+  </servlet>
+
+  <servlet-mapping>
+	<servlet-name>discoverer_add_offering</servlet-name>
+	<url-pattern>/addoffer</url-pattern>
+  </servlet-mapping>
+
```
mbarrientos: @buccarel I see you created several servlets but you're only exposing `/addoffer`. Is this intentional?

mbarrientos:
@buccarel Thanks for the PR. I see that you are using low level servlets to create the web services. Have you considered using some JAX-RS implementation, like `Jersey`? Or using it together with some framework such `spring-boot` or `dropwizard`. 

If it helps, you can have a look to how the resources and services are exposed at `dashboard`: https://github.com/SeaCloudsEU/SeaCloudsPlatform/blob/master/dashboard/src/main/java/eu/seaclouds/platform/dashboard/resources/DeployerResource.java#L46

This is just a recommendation, i'd be OK to accept servlets as far as they do their job.

buccarel:
@mbarrientos Sorry for the fake description, I forgot to change it. Imma do it immediately.

Yep, for the moment, I intentionally exposed the add_offer only because 1) I need to sync with other partners about the I/O behavior of the several functionalities; 2) the add_offer is the one implementing the manual insertion of the offerings into the repository.

I studied the Jersey implementation for the servlet and I saw that there is no much difference w.r.t. what I did (i.e., instead of a doPost, you have to mark a method with @Post, a servlet container like tomcat is still required, ecc...). Since I had some template already written for servlets like that, I used that to save time.

mbarrientos:
The JSON/XML (de)serialization support in `Jersey` is also very valuable, even more when we have a common object model. 

As a general comment, we should try to merge just the features that are implemented. If the rest of the servlets are not being used, or they're not yet implemented (like https://github.com/SeaCloudsEU/SeaCloudsPlatform/pull/72/files#diff-b9ba47ff6b4da9cc8d939622ce2b229dR43) they can be added in another PR. 

However, it generally looks good, +1.

buccarel:
@szenzaro @mbarrientos Ok I did the fixed we discussed above and I squashed all the commits again. If it's +2, it's ready to be accepted, I guess.

perezp:
For the next times, we should take into account the comment from @mbarrientos regarding PRs including the skeleton of a feature but not the implementation.

PR #73, opened 2015-07-15
Added SLA Core into SeaClouds Platform

rosogon:


MicheleGuerriero:
I had a look at this PR and to me it seems ok. I didn't read all the code into details even because as far as I understand this PR is just to move the SLA code into the SeaClouds Platform repository but any real changes is introduced. So to me it can be merged, but I think is better to wait for more expert opinions.

rosogon:
@mbarrientos , @adriannieto : As Michele says, there are no new features, so maybe the review needs to be done on the pom side, and other project standards. Could you have a look at it?

mbarrientos:
@rosogon Probably there's no need on creating two maven parents `sla` and `sla-core` to avoid confusion. I'd suggest to remove `sla` parent pom and leave just `sla-core`. If there's any plan on creating more siblings  for sla you could keep the sla folder without need of it being a maven module.

rosogon:
There is a plan to add a sibling for sla-core (not for sla): the sla-generator. That's why the parent sla is there. The sla-generator is outside the sla-core, and I don't like to place it on root.

mbarrientos:
My suggestion was to leave `sla/` as a non-maven directory level (same way as `usage/`) but I'm okay with this too. +1

PR #74, opened 2015-07-15
FEATURE: Adding CloudFoundry support to the deployer.

kiuby88:
The deployer uses a custom Brooklyn distribution which contains the CloudFoundry support.

The version of Brooklyn which is being used to deploy SeaCloudsPlatform was uploaded to `0.8.0-SNAPSHOT`. Previously, we used a fix SNAPSHOT version of `0.7.0`, so probably we should to fix a `0.8.0` version but currently there is not any available at any of the Brooklyn snapshot repositories.

Currently, the Deployer is using a custom distribution of Brooklyn which contains the CloudFoundry support (entities and location). This [custom version](https://github.com/kiuby88/incubator-brooklyn/releases) is stored in my SeaClouds/incubator-brooklyn fork. It is a temporal solution, we are working on the total integration of CloudFoundry features into [apache/incubator-brooklyn](https://github.com/apache/incubator-brooklyn).

Thanks @mbarrientos for the support.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -94,7 +94,7 @@
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
         <guava.version>18.0</guava.version>
         <javatuples.version>1.2</javatuples.version>
-        <brooklyn.version>0.7.0-20150614.2158</brooklyn.version>
+        <brooklyn.version>0.8.0-SNAPSHOT</brooklyn.version>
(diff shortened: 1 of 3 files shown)


andreaturli:
I strongly believe this is not what we want to merge. @kiuby88 you should first move your code from your repo to seacloudsPlatform and then re-open this PR if still needed.

kiuby88:
I agree. This PR will be closed

PR #75, opened 2015-07-17
FIX(Dashboard): fixed wrong artifact bug

mbarrientos:
Fixed bug where dashboard brooklyn entity was downloading the wrong artifact. 
Also setting default values for config keys.


mbarrientos:
@rosogon @MicheleGuerriero @perezp Could any of you guys have a look? Thanks!

rosogon:
@mbarrientos : I don't understand the purpose of this file. Does it "receive" the values specified in the blueprint and set default values if not set in the blueprint?

mbarrientos:
@rosogon Thanks for the review. 

The config keys default values are only used if there's nothing specified in the blueprint. In the way that the dashboard is built right now, all configuration values are mandatory so having default values gives a bit more consistency on built.

rosogon:
+1

perezp:
+1

PR #78, opened 2015-07-17
FIX(usage): Added T4C env vars to sla core

rosogon:
These lines were missing from the master.  

I'm not sure if they are still valid. F.e., deployer uses $brooklyn:component("monitoring").component("child","manager"), instead of directly component("manager").


rosogon:
@adriannieto : Can you review this?

adriannieto:
@rosogon  could you check this?

rosogon:
Changed vars as suggested by Adrián. SLA receives right values.

andreaturli:
I don't have tested the deployment but the changes look sensible.

lgtm +1

PR #79, opened 2015-07-20
FEATURE(common) Serializer: Read TOSCA from string

szenzaro:
TOSCA is now readable also from strings.
@perezp can you check this?


perezp:
Than you Simone. 
+1 
I've seen that there are some problems with the logger when "common" runs its tests. We should take care of them in next PRs, not in this one. 

Are you waiting for comments from more developers or can I merge it now?

szenzaro:
I am aware of the conflicts with the loggers (it seems that there are two providing the same logger), but I do not know if it is a local problem or a project setting configuration to be changed.

In my opinion we can merge.

PR #80, opened 2015-07-20
FIX(common) Cleanup of the object model

szenzaro:
Since we have decided to not use the OM, I think we can remove it from the repository to keep it cleaner.
@andreaturli can you review this?


andreaturli:
@szenzaro I think it is a good clean up. As travis is happy with this removal, so do I.

+1

PR #81, opened 2015-07-20
Major refactoring for dashboard/api and common

mbarrientos:
Major refactoring of the dashboard and common modules. 
- Common is now divided into "entities" + "model" + "utils".
- Dashboard is isolated so that it only includes the JS project.
- SeaClouds REST API definition has been moved to /usage/rest-api (previously inside Dashboard)
- Dashboard backend (Dropwizard app) was moved to /usage/server
- Dependencies management have been pulled up to parent from all modules.


szenzaro:
I think the commits should be squashed before merging them.

andreaturli:
@mbarrientos does this PR needs to touch so many areas of the project? I think it is hard to properly review a PR like that, I'd prefer to have incremental PRs rather than a big one, particularly because it deals with many modules. Wdyt?

mbarrientos:
@andreaturli I agree it touches too many modules, but there is a lot of deleting JS libs and moving files from one folder to the other. I also couldn't isolate some features, like converting dashboard to JS project and creating a rest-api module, since they're related. 

Do you suggest any action? I could provide further documentation on the changes if needed (preferable to split features).

andreaturli:
not sure I understand why 'common.entites' renaming should be part of this PR, for example.
I think we agreed on having self-contained PR, specific to one topic only. No strong feelings, though, so not sure I have a good suggestion :(

mbarrientos:
Since I didn't receive any other comment for this PR, if you guys agree i'll just close this and try to split features a bit more in the subsequent ones.

adriannieto:
@andreaturli the problem is isolating this into a for example 3 PR it's going to be a tough task because for example if we move the API from dashboard to rest-api we will break in this PR the master tree until the remaining  PR will be accepted. We will take a look to study further what could be removed without breaking the master tree

andreaturli:
@adriannieto my comment was mostly focusing on the type of the PR. I made a very general comment that is valid for every PR. My suggestion is to have small PR unless really needed!
I don't know anything about this rest-api idea, of course I cannot discuss the details of every single line of code! :)

Fast iterations are to be preferred. If the dashboard needs a lot of refactoring, better plan an epic composed of multiple backlog item which will be addressed on different PRs, don't you think?
Another reason to avoid very large PRs is that they are likely to touch a lot of modules/areas so they will break/block other developers for long time.

Thoughts?

mbarrientos:
Part of this PR is being covered on #83, so closing.

PR #82, opened 2015-07-20
Feature/cf paas support

kiuby88:
CloudFoundry support is added.

CloudFoundry location was added which allows to manage the services provided by CloudFoundry.
A set of entities represent the Applications which can be deployed on CF and the services that can be instantiated and bound to the application.

The CloudFoundry (CF) entities and location are integrated on SeaCloudsPlatform.
A custom jar is generated for Deployer module which contains the required dependencies to manage the CF location and entities.

For executing the (Live) tests, you need to add the CloudFoundry location definition to the local `brooklyn.properties`, of course you need an enable count in a CF provider, currently I recommend [Pivotal Web Services](https://run.pivotal.io/) (as we can see in the following example).

```
## Cloud Foundry
brooklyn.location.named.cloudfoundry-instance = cloudfoundry
brooklyn.location.named.cloudfoundry-instance.user = an-email
brooklyn.location.named.cloudfoundry-instance.password = a-pass
brooklyn.location.named.cloudfoundry-instance.org = an-org-name
brooklyn.location.named.cloudfoundry-instance.endpoint = https://api.run.pivotal.io
brooklyn.location.named.cloudfoundry-instance.space = development
brooklyn.location.named.cloudfoundry-instance.address = run.pivotal.io
```

TODO:
* Optimice the Deployer.jar building, currently is very big.
* Add User Service (which were develop by @mbarrientos).
* Add OpenShift support (first approach is being enhanced). 
* Change the packeges name to `eu.seaclouds...`
* Adding Autoscaling support


deployer/pom.xml
```diff
@@ -16,23 +16,118 @@
...
+        <dependency>
+            <groupId>org.apache.brooklyn</groupId>
+            <artifactId>brooklyn-software-base</artifactId>
+            <version>${brooklyn.version}</version>
+        </dependency>
+
+        <dependency>
+            <groupId>org.cloudfoundry</groupId>
+            <artifactId>cloudfoundry-client-lib</artifactId>
+            <version>1.1.2</version>
```
andreaturli: create a <cloudfoundry-client-lib.version> property into pom.xml

deployer/pom.xml
```diff
@@ -16,23 +16,118 @@
...
+        </dependency>
+
+        <dependency>
+            <groupId>mysql</groupId>
+            <artifactId>mysql-connector-java</artifactId>
+            <version>5.1.35</version>
```
andreaturli: do you need this dep? if yes, create a property for the version

deployer/pom.xml
```diff
@@ -16,23 +16,118 @@
...
+            </plugin>
 
+            <plugin>
+                <artifactId>maven-jar-plugin</artifactId>
+                <version>2.3.1</version>
+                <executions>
+                    <execution>
+                        <id>default-jar</id>
+                        <!-- put the default-jar in the none phase to skip it from being created -->
+                        <phase>none</phase>
```
andreaturli: could you explain this setting?

deployer/src/main/assembly/assembly.xml
```diff
@@ -0,0 +1,49 @@
...
+            <excludes>
+                <exclude>org.apache.brooklyn:brooklyn-core</exclude>
+                <exclude>org.apache.brooklyn:brooklyn-software-base</exclude>
+                <exclude>org.apache.brooklyn:brooklyn-logback-xml</exclude>
+                <exclude>org.slf4j:slf4j-api</exclude>
+                <exclude>ch.qos.logback:logback-classic</exclude>
+            </excludes>
+        </dependencySet>
+    </dependencySets>
+</assembly>
```
andreaturli: add an empty line

deployer/src/test/resources/phpHelloWorld/index.php
```diff
@@ -0,0 +1,26 @@
...
+  Unless required by applicable law or agreed to in writing,
+  software distributed under the License is distributed on an
+  "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+  KIND, either express or implied.  See the License for the
+  specific language governing permissions and limitations
+  under the License.
+ -->
+<html>
+ <head>
+  <title>Pruebsa de PHP</title>
```
andreaturli: better in English?

deployer/src/test/java/eu/seaclouds/location/cloudfoundry/ServiceCredentials.java
```diff
@@ -0,0 +1,112 @@
...
+import org.testng.annotations.DataProvider;
+import org.testng.annotations.Test;
+
+import java.util.ArrayList;
+import java.util.List;
+
+import static org.testng.Assert.assertNotNull;
+import static org.testng.Assert.assertTrue;
+
+public class ServiceCredentials extends AbstractCloudFoundryPaasLocationLiveTest  {
```
andreaturli: rename it to ServiceCredentialLiveTest

deployer/src/test/java/eu/seaclouds/location/cloudfoundry/JavaWebAndServiceCloudFoundryLiveTest.java
```diff
@@ -0,0 +1,123 @@
...
+
+    private final String SQL_ARTIFACT_URL = getClasspathUrlForResource(SQL_ARTIFACT_NAME);
+    private final String APPLICATION_ARTIFACT_URL =
+            getClasspathUrlForResource(APPLICATION_ARTIFACT_NAME);
+
+    private final String SERVICE_NAME = APPLICATION_SERVICE_NAME+"-mysql";
+    private final String SERVICE_TYPE_ID = "cleardb";
+    private final String SERVICE_PLAN = "spark";
+
+    @Test(groups = {"Live"})
```
andreaturli: move @Test definition at class level

deployer/src/test/java/eu/seaclouds/location/cloudfoundry/JavaWebAndServiceCloudFoundryLiveTest.java
```diff
@@ -0,0 +1,123 @@
...
+                assertEquals(server.getAttribute(JavaCloudFoundryPaasWebApp.INSTANCES_NUM),
+                        PaasHardwareResources.REQUIRED_INSTANCES.getDefaultValue());
+                assertEquals(server.getAttribute(JavaCloudFoundryPaasWebApp.MEMORY),
+                        PaasHardwareResources.REQUIRED_MEMORY.getDefaultValue());
+            }
+        });
+    }
+
+    @Test(expectedExceptions = brooklyn.util.exceptions.PropagatedRuntimeException.class,
+    groups={"Live"})
```
andreaturli: move @Test definition at class level

deployer/src/test/java/eu/seaclouds/location/cloudfoundry/CloudFoundryYamlTest.java
```diff
@@ -0,0 +1,134 @@
...
+    private final String SERVER_NAME = "test-brooklyn-app-from-yaml";
+    private final String JDBC_SENSOR = SERVER_NAME+".credentials.jdbcUrl";
+    private final String NAME_SENSOR = SERVER_NAME+".credentials.name";
+    private final String HOSTNAME_SENSOR = SERVER_NAME+".credentials.hostname";
+    private final String USERNAME_SENSOR = SERVER_NAME+".credentials.username";
+    private final String PASSWORD_SENSOR = SERVER_NAME+".credentials.password";
+    private final String PORT_SENSOR = SERVER_NAME+".credentials.port";
+
+
+    @Test( groups={"Live"} )
```
andreaturli: move @Test definition at class level

deployer/src/test/java/eu/seaclouds/location/cloudfoundry/CloudFoundryYamlTest.java
```diff
@@ -0,0 +1,134 @@
...
+import brooklyn.test.Asserts;
+import brooklyn.util.text.Strings;
+import org.testng.annotations.Test;
+
+import static org.testng.Assert.assertEquals;
+import static org.testng.Assert.assertFalse;
+import static org.testng.Assert.assertNotNull;
+import static org.testng.Assert.assertTrue;
+
+public class CloudFoundryYamlTest {
```
andreaturli: rename it CloudFoundryYamlLiveTest

deployer/src/test/java/eu/seaclouds/location/cloudfoundry/CloudFoundryPaasLocationResolverTest.java
```diff
@@ -0,0 +1,92 @@
...
+    private BrooklynProperties brooklynProperties;
+
+    private final String USER = "user";
+    private final String PASSWORD = "password";
+    private final String ORG = "organization";
+    private final String SPACE = "space";
+    private final String ENDPOINT = "endpoint";
+    private final String ADDRESS = "run.pivotal.io";
+
+    @BeforeMethod(alwaysRun = true)
```
andreaturli: this is the default, remove it

deployer/src/test/java/eu/seaclouds/location/cloudfoundry/CloudFoundryPaasLocationResolverTest.java
```diff
@@ -0,0 +1,92 @@
...
+        brooklynProperties.put("brooklyn.location.named.cloudfoundry-instance", "cloudfoundry");
+        brooklynProperties.put("brooklyn.location.named.cloudfoundry-instance.user", USER);
+        brooklynProperties.put("brooklyn.location.named.cloudfoundry-instance.password", PASSWORD);
+        brooklynProperties.put("brooklyn.location.named.cloudfoundry-instance.org", ORG);
+        brooklynProperties.put("brooklyn.location.named.cloudfoundry-instance.endpoint", ENDPOINT);
+        brooklynProperties.put("brooklyn.location.named.cloudfoundry-instance.space", SPACE);
+        brooklynProperties.put("brooklyn.location.named.cloudfoundry-instance.address", ADDRESS);
+    }
+
+    @AfterMethod(alwaysRun = true)
```
andreaturli: this is the default, remove it

deployer/src/test/java/eu/seaclouds/location/cloudfoundry/CloudFoundryPaasLocationResolverTest.java
```diff
@@ -0,0 +1,92 @@
...
+    public void tearDown() throws Exception {
+        if (managementContext != null) {
+            managementContext.terminate();
+        }
+    }
+
+    @Test
```
andreaturli: move annotation to class, remove it from methods

deployer/src/test/java/eu/seaclouds/location/cloudfoundry/CloudFoundryPaasLocationLiveTest.java
```diff
@@ -0,0 +1,45 @@
...
+
+import org.cloudfoundry.client.lib.CloudFoundryClient;
+import org.testng.annotations.Test;
+
+import static org.testng.Assert.assertEquals;
+import static org.testng.Assert.assertNotNull;
+
+public class CloudFoundryPaasLocationLiveTest extends AbstractCloudFoundryPaasLocationLiveTest {
+
+    @Test(groups = {"Live"})
```
andreaturli: move it to the class, remove them from methods

deployer/src/test/java/eu/seaclouds/location/cloudfoundry/CloudFoundryClearDbServiceLiveTest.java
```diff
@@ -0,0 +1,60 @@
...
+import brooklyn.test.Asserts;
+import com.google.common.collect.ImmutableList;
+import org.testng.annotations.Test;
+
+import static org.testng.Assert.assertEquals;
+import static org.testng.Assert.assertTrue;
+
+public class CloudFoundryClearDbServiceLiveTest extends AbstractCloudFoundryPaasLocationLiveTest  {
+
+    @Test(groups = {"Live"})
```
andreaturli: remove annotation, move it to class def

deployer/src/test/java/eu/seaclouds/location/cloudfoundry/AbstractCloudFoundryPaasLocationLiveTest.java
```diff
@@ -0,0 +1,76 @@
...
+    protected BrooklynProperties brooklynProperties;
+    protected LocalManagementContext managementContext;
+    protected CloudFoundryPaasLocation cloudFoundryPaasLocation;
+    protected final String LOCATION_SPEC_NAME = "cloudfoundry-instance";
+    protected final String APPLICATION_NAME = "test-brooklyn-application-" + UUID.randomUUID()
+            .toString().substring(0, 8);
+    protected final String APPLICATION_SERVICE_NAME = "test-brooklyn-service-" + UUID.randomUUID()
+            .toString().substring(0, 8);
+
+    @BeforeMethod(alwaysRun = true)
```
andreaturli: remove it

deployer/src/test/java/eu/seaclouds/location/cloudfoundry/AbstractCloudFoundryPaasLocationLiveTest.java
```diff
@@ -0,0 +1,76 @@
...
+        brooklynProperties = new LocalManagementContext().getBrooklynProperties();
+        cloudFoundryPaasLocation = newSampleCloudFoundryLocationForTesting(LOCATION_SPEC_NAME);
+        app = TestApplication.Factory.newManagedInstanceForTests();
+    }
+
+    protected LocalManagementContext newLocalManagementContext() {
+        return new LocalManagementContextForTests(BrooklynProperties.Factory.newDefault());
+    }
+
+    @AfterMethod(alwaysRun = true)
```
andreaturli: remove it

deployer/pom.xml
```diff
@@ -16,23 +16,118 @@
...
+        </dependency>
+    </dependencies>
 
-  <properties>
-    <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
-  </properties>
+    <build>
+        <plugins>
+            <plugin>
+                <artifactId>maven-assembly-plugin</artifactId>
```
andreaturli: do we really need to create an assembly here?

deployer/src/main/java/brooklyn/location/cloudfoundry/CloudFoundryPaasLocation.java
```diff
@@ -0,0 +1,85 @@
...
+import org.slf4j.LoggerFactory;
+
+import java.net.MalformedURLException;
+import java.net.URI;
+import java.net.URL;
+
+public class CloudFoundryPaasLocation extends AbstractLocation
+        implements PaasLocation, PaasHardwareResources {
+
+    public static final Logger LOG = LoggerFactory.getLogger(CloudFoundryPaasLocation.class);
```
andreaturli: this doesn't seem to be used

deployer/src/main/java/brooklyn/location/cloudfoundry/CloudFoundryPaasLocation.java
```diff
@@ -0,0 +1,85 @@
...
+            client = new CloudFoundryClient(credentials,
+                    getTargetURL(getConfig(CF_ENDPOINT)),
+                    getConfig(CF_ORG), getConfig(CF_SPACE), true);
+            client.login();
+        }
+    }
+
+    @Override
+    public String getPaasProviderName() {
+        return "CloudFoundry";
```
andreaturli: this looks like the name of an API rather than the Provider name, I'd expect Pivotal as provider name. Thoughts?

deployer/src/main/java/brooklyn/location/cloudfoundry/CloudFoundryPaasLocationResolver.java
```diff
@@ -0,0 +1,74 @@
...
+    public String getPrefix() {
+        return CLOUD_FOUNDRY;
+    }
+
+    @Override
+    public boolean accepts(String spec, LocationRegistry registry) {
+        if (BasicLocationRegistry.isResolverPrefixForSpec(this, spec, true)) {
+            return true;
+        } else {
+            // TODO: check valid CloudFoundry format on spec
```
andreaturli: maybe `return BasicLocationRegistry.isResolverPrefixForSpec(this, spec, true)` for now?

deployer/src/main/java/brooklyn/location/cloudfoundry/CloudFoundryPaasLocationResolver.java
```diff
@@ -0,0 +1,74 @@
...
+        }
+        return false;
+    }
+
+    @Override
+    @SuppressWarnings("unchecked")
+    public Location newLocationFromString(Map locationFlags,
+                                          String spec,
+                                          LocationRegistry registry) {
+        // TODO: TODO
```
andreaturli: what else needs to be done: please improve TODO

deployer/src/main/java/brooklyn/location/cloudfoundry/PaasHardwareResources.java
```diff
@@ -0,0 +1,52 @@
...
+ */
+package brooklyn.location.cloudfoundry;
+
+import brooklyn.config.ConfigKey;
+import brooklyn.entity.basic.ConfigKeys;
+import brooklyn.event.AttributeSensor;
+import brooklyn.event.basic.Sensors;
+import brooklyn.util.flags.SetFromFlag;
+
+public interface PaasHardwareResources {
```
andreaturli: Would it be better to call it `PaasLocationConfig` for simmetry with `JcloudsLocationConfig`?

deployer/src/main/java/brooklyn/entity/cloudfoundry/PaasEntityDriver.java
```diff
@@ -0,0 +1,64 @@
...
+
+import brooklyn.entity.basic.EntityLocal;
+import brooklyn.entity.drivers.EntityDriver;
+
+public interface PaasEntityDriver extends EntityDriver {
+
+    /**
+     * The entity whose components we are controlling.
+     */
+    EntityLocal getEntity();
```
andreaturli: I think this is defined in the EntityDriver already

deployer/src/main/java/brooklyn/entity/cloudfoundry/CloudFoundryEntityImpl.java
```diff
@@ -0,0 +1,568 @@
...
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package brooklyn.entity.cloudfoundry;
+
+
```
andreaturli: remove this empty line

deployer/src/main/java/brooklyn/entity/cloudfoundry/CloudFoundryEntity.java
```diff
@@ -0,0 +1,62 @@
...
+import brooklyn.entity.proxying.ImplementedBy;
+import brooklyn.entity.trait.Startable;
+import brooklyn.event.AttributeSensor;
+import brooklyn.event.basic.Sensors;
+import brooklyn.location.cloudfoundry.CloudFoundryPaasLocation;
+import brooklyn.util.flags.SetFromFlag;
+import brooklyn.util.time.Duration;
+
+@ImplementedBy(CloudFoundryEntityImpl.class)
+public interface CloudFoundryEntity extends Entity, Startable {
```
andreaturli: I suspect this new entity hierarchy is not really needed, isn't it?

I think we agreed on trying to have a design that reuse the same entity definition and using the driver (the only changing part) the deployer will be able to deploy on IaaS or PaaS?

I appreciate `newDriver` requires a `MachineLocation` which you don't have as `PaasLocation` doesn't extend it, maybe we should patch brooklyn code to accept a more generic AbstractLocation instead, rather than adding this new duplicate hierarchy of entities. wdyt?

deployer/src/main/java/brooklyn/entity/cloudfoundry/webapp/CloudFoundryWebAppImpl.java
```diff
@@ -0,0 +1,253 @@
...
+import brooklyn.util.time.Duration;
+import com.google.common.base.Functions;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.util.LinkedList;
+import java.util.Map;
+import java.util.concurrent.Callable;
+
+public abstract class CloudFoundryWebAppImpl extends CloudFoundryEntityImpl
```
andreaturli: I think this should extend `SoftwareProcessImpl` rather than `CloudFoundryEntityImpl`?
What happens when we want to support PaaS X, we need to duplicate over and over PaaSXEntityImpl? We need to improve this design!

andreaturli:
Thanks @kiuby88, very good start!
My initial review focused on maven structure and `location.cloudfoundry` package
I think we don't need to create an assembly out of this maven module, please confirm.
The cloudfoundry location is a good start, just minor comments.

Re `entity.cloudfoundry` there are more concerns:
- package name should be `cloudfoundry.entity` maybe?
But before that, do we really need this new hierarchy of entities: I think the ideal design would be to have the same entity definition reusable on different env (IaaS / PaaS etc) using different drivers. Although not difficult, it doesn't feel right to duplicate the definitiion of the same entity for different env.
Could you comment on this?

I haven't looked at CloudFoundryService idea also. I'll do when the previous comments are addressed.

Thanks again!

kiuby88:
Thanks a lot Andrea, and thanks @mbarrientos for the support.

About the assembly, as you can see, the `deployer-0.8.0-SNAPSHOT.jar` is added to the `BrooklynNode` classpath in order to add the CloudFoundry entities and location, but this jar needs some dependencies for working fine, for example the `cloudfoundry-client-lib`. Then, I use an assembly which builds a jar which contains the basic required dependencies as `mysql` or `cf-client` and I have excluded brooklyn dependencies because they will be provided by `BrooklynNode`.

About the packaging, we are adding the first entities for CF as a temporal solution, while @mbarrientos  and I are working in brooklyn for adding PaaS to current entities.

thx

deployer/pom.xml
kiuby88: Yes, ClearDbService needs this dependency. I will create a version in the parent.

deployer/pom.xml
kiuby88: I want to create the `deployer-0.8.0-SNAPSHOT.jar` using the assembly, so I have avoided the jar for default using this plugin.

deployer/src/test/resources/phpHelloWorld/index.php
kiuby88: Of course

deployer/pom.xml
andreaturli: As a temporary solution, while we wait to merge CloudFoundry to Apache Brooklyn, I'd suggest you to add the needed dependencies in the yaml classpath, so that the `location/cloudfoundry` will be ready to be merged to brooklyn. Makes sense?

deployer/src/main/java/brooklyn/location/cloudfoundry/CloudFoundryPaasLocation.java
kiuby88: Yes. We used during developing, but it will be able to remove.

deployer/src/main/java/brooklyn/location/cloudfoundry/CloudFoundryPaasLocation.java
kiuby88: It is an important issue :+1: 
I think we spoke about this before.

We started using Pivotal, but we decided to move to cloudfoundry because it allows to represent any CloudFoundry provider. We could change the name of the method to `getPaasApiName` or similar. However, it is not an isolated change because other PaasLocations, such as OpenShiftLocation, should override this method and in those cases the method `getPaasProviderName` makes sense. 

We could use something like `getPaasProviderIdentifier` or similar and avoid a concrete aspect (name, or API) of any provider.

WDYT?
As I mentioned, I think it is an important point

deployer/src/main/java/brooklyn/location/cloudfoundry/CloudFoundryPaasLocationResolver.java
kiuby88: Really, if we do not want to check anything, the methods could be something as following:
```
@Override
    public boolean accepts(String spec, LocationRegistry registry) {
        return BasicLocationRegistry.isResolverPrefixForSpec(this, spec, true);
    }

right?
```

deployer/src/main/java/brooklyn/location/cloudfoundry/CloudFoundryPaasLocationResolver.java
kiuby88: The error was mine

deployer/src/main/java/brooklyn/location/cloudfoundry/PaasHardwareResources.java
kiuby88: Good point of view

deployer/src/main/java/brooklyn/entity/cloudfoundry/PaasEntityDriver.java
kiuby88: yes, I will remove it

deployer/src/main/java/brooklyn/entity/cloudfoundry/webapp/CloudFoundryWebAppImpl.java
kiuby88: Yes, as you have mentioned above, this adds a new hierarchy. But we have added it as a temporal solution until @mbarrientos and I will integrate `PaasLocation` in the current Brooklyn entities.

deployer/src/main/java/brooklyn/location/cloudfoundry/CloudFoundryPaasLocation.java
andreaturli: leave it as-is for now, we will need to change it soon

deployer/src/main/java/brooklyn/location/cloudfoundry/CloudFoundryPaasLocationResolver.java
andreaturli: yeah

deployer/src/main/java/brooklyn/entity/cloudfoundry/webapp/CloudFoundryWebAppImpl.java
andreaturli: I'd prefer to have this first commit made right, rather than clean up things later on. Thoughts?

deployer/pom.xml
kiuby88: I was my first idea. But, if we build a jar with any dependency we will have to add these dependencies to the YAML, and it could be an error prone task because they are a lot of dependencies. I think this new assembly it is not a problem for Brooklyn merging because we are refactoring the CF entities in Brooklyn.

Of course it is just my experience and if you know a better solution we are able to check it.

deployer/src/main/java/brooklyn/location/cloudfoundry/CloudFoundryPaasLocation.java
kiuby88: :+1:

deployer/pom.xml
andreaturli: you'll provide the yaml blueprint anyway, so it is not error prone, they won't change too much

deployer/src/main/java/brooklyn/entity/cloudfoundry/CloudFoundryEntity.java
kiuby88: Precisely, we are working on this in Brooklyn. We are trying to add more genericity to this class.

deployer/src/main/java/brooklyn/entity/cloudfoundry/webapp/CloudFoundryWebAppImpl.java
kiuby88: I am not sure. This first commit contains a first approach our work about CF integration in SeaClouds. As I mentioned previously @mbarrientos and I are working in the CF integration in Brooklyn. But I think it will be a large process because we are changing several aspect that should be discuss.
I'd prefer adding this commit and in the future changing the necessary entities.

deployer/pom.xml
kiuby88: I can try it.

kiuby88: Currently we have added `deployer-0.8.0-SNAPSHOT.jar` as a  resource in `usage/installer/pom.xml`, as we can see following:

```
<resources>
    <resource>
        <directory>${basedir}/../../deployer/target</directory>
        <includes>
                <include>deployer-0.8.0-SNAPSHOT.jar</include>
        </includes>
    </resource>
        ...
</resources>
```
It allows to Brooklyn host ( which deploys `SeaCloudsPlatform` ) finds the `deployer-0.8.0-SNAPSHOT.jar` from the classpath and adds it to `BrooklynNode`. This jar ,that is generated by the assembly, includes all `cloudfoundry` and `mysql` dependencies. 

If we do not include the jar as a resource into `installer-0.8.0-SNAPSHOT.jar`, Brooklyn host will not find `deployer-0.8.0-SNAPSHOT.jar` from the classpath. 
We tried to add `deployer-0.8.0-SNAPSHOT.jar` to the classpath from maven, using a dependency, and the jar is added to the classpath but the jar is not found from the classpath.   because when Brooklyn receives a URL based on classpath, as following:
```
brooklynnode.classpath:
      - classpath://deployer-0.8.0-SNAPSHOT.jar
```
it tries to find this jar as a resource using `gerLocader().getResurce("deployer-0.8.0-SNAPSHOT.jar")`, however the `deployer-0.8.0-SNAPSHOT.jar` was not added as a resource, (it was added as maven dependency).

Then, if we do not want to build the custom `deployer-0.8.0-SNAPSHOT.jar` using the assembly, we should add the needed dependencies as resources too. It could be added to  `installer-0.8.0-SNAPSHOT.jar`, it was my first idea, however these dependencies (and the transitive dependencies) should be adde to `installer/pom.xml` from `.m2` repositories, something like following:
```
<resources>
    <resource>
        <directory>${user.home}.m2/repository/org/cloudfoundry/cloudfoundry-client-lib/1.1.2</directory>
         <includes>
             <include>cloudfoundry-client-lib-1.1.2</include>
         </includes>
    </resource>
</resources>
```
Next, this resource (jar dependencies) should to be added to `BrooklynNode` from the classpath.

I have developed this approach for checking if it works fine.

kiuby88: @andreaturli  I have deleted the deployer assembly and I have added all dependencies as resources to installer/pom.xml.
As I have mentioned above, I think that maintain this approach will be an error prone task, because we will have to maintain the integrity between the different dependencies versions if we modified any version.
If we add a new dependency, such as OpenShift client, we will add this dependency and its transitive dependencies as resource and to` BrooklyNode` classpath.

Moreover, I think that the number of dependencies that we need to add to the BrooklynNode classpath is too big.
Then, from my point of view I think it is easier the deployer assembly. However, it is just my opinion, the approach of this version works correctly so we could be to use it. Thoughts?

andreaturli: I see your point, although not a big fan, I think you should consider
maven-shade-plugin to create a uber jar, with all the necessary includes.
This the module will produce a simple jar and a uber-jar to be used in the
yaml blueprint. Like it?

Il giorno mar 21 lug 2015 19:07 Jose <notifications@github.com> ha scritto:

> In deployer/pom.xml
> <https://github.com/SeaCloudsEU/SeaCloudsPlatform/pull/82#discussion_r35127329>
> :
>
> >
> > +            <plugin>
> > +                <artifactId>maven-jar-plugin</artifactId>
> > +                <version>2.3.1</version>
> > +                <executions>
> > +                    <execution>
> > +                        <id>default-jar</id>
> > +                        
> > +                        <phase>none</phase>
>
> @andreaturli <https://github.com/andreaturli> I have deleted the deployer
> assembly and I have added all dependencies as resources to
> installer/pom.xml.
> As I have mentioned above, I think that maintain this approach will be an
> error prone task, because we will have to maintain the integrity between
> the different dependencies versions if we modified any version.
> If we add a new dependency, such as OpenShift client, we will add this
> dependency and its transitive dependencies as resource and toBrooklyNode
> classpath.
>
> Moreover, I think that the number of dependencies that we need to add to
> the BrooklynNode classpath is too big.
> Then, from my point of view I think it is easier the deployer assembly.
> However, it is just my opinion, the approach of this version works
> correctly so we could be to use it. Thoughts?
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/SeaCloudsEU/SeaCloudsPlatform/pull/82/files#r35127329>
> .
>

-- 
Cloudsoft Corporation Limited, Registered in Scotland No: SC349230. 
 Registered Office: 13 Dryden Place, Edinburgh, EH9 1RP
 
This e-mail message is confidential and for use by the addressee only. If 
the message is received by anyone other than the addressee, please return 
the message to the sender by replying to it and then delete the message 
from your computer. Internet e-mails are not necessarily secure. Cloudsoft 
Corporation Limited does not accept responsibility for changes made to this 
message after it was sent.

Whilst all reasonable care has been taken to avoid the transmission of 
viruses, it is the responsibility of the recipient to ensure that the 
onward transmission, opening or use of this message and any attachments 
will not adversely affect its systems or data. No responsibility is 
accepted by Cloudsoft Corporation Limited in this regard and the recipient 
should carry out such virus and other checks as it considers appropriate.

kiuby88: Thanks for your reply. I took a look to this plugin yesterday but I thought it was not useful. I will add  it, it could be a good solution.
Thx again.

kiuby88: I have added share plugin. I found a problem with .SF and .DSA, so I excluded them.

andreaturli:
@kiuby88 travis is not happy, can you see why?

kiuby88:
Yes, I should to rebase the branch

deployer/pom.xml
```diff
@@ -16,23 +16,119 @@
...
+    <packaging>jar</packaging>
+
+    <name>deployer</name>
+    <url>http://maven.apache.org</url>
+
+    <dependencies>
+        <dependency>
+            <groupId>org.apache.brooklyn</groupId>
+            <artifactId>brooklyn-core</artifactId>
+            <version>${brooklyn.version}</version>
```
andreaturli: I think this could be `<scope>provided</scope>`

deployer/pom.xml
```diff
@@ -16,23 +16,119 @@
...
+        </dependency>
+
+        <dependency>
+            <groupId>org.apache.brooklyn</groupId>
+            <artifactId>brooklyn-software-base</artifactId>
+            <version>${brooklyn.version}</version>
```
andreaturli: I think this could be `<scope>provided</scope>`

deployer/pom.xml
```diff
@@ -16,23 +16,119 @@
...
+                 provided so that it isn't put into the assembly (as it supplies its own explicit logback.xml);
+                 see Logging in the Brooklyn website/userguide for more info -->
+            <optional>true</optional>
+            <scope>provided</scope>
+        </dependency>
+    </dependencies>
 
+    <build>
+        <plugins>
+            <plugin>
```
andreaturli: If you mark brooklyn dep as `provided` (see pom.xml's suggestions) you could have

```
            <plugin>
                <groupId>org.apache.maven.plugins</groupId>
                <artifactId>maven-shade-plugin</artifactId>
                <version>2.4.1</version>
                <configuration>
                    <createDependencyReducedPom>false</createDependencyReducedPom>
                    <filters>
                        <filter>
                            <artifact>*:*</artifact>
                            <excludes>
                                <exclude>META-INF/*.SF</exclude>
                                <exclude>META-INF/*.DSA</exclude>
                            </excludes>
                        </filter>
                    </filters>
                </configuration>
                <executions>
                    <execution>
                        <phase>package</phase>
                        <goals>
                            <goal>shade</goal>
                        </goals>
                    </execution>
                </executions>
            </plugin>
```
which brings the uber-jar from 63 MB down to 18 MB. wdyt?
Can you test if that uber-jar still works?

deployer/src/test/java/eu/seaclouds/location/cloudfoundry/CloudFoundryPaasLocationResolverTest.java
kiuby88: But, if I remove the annotation, this method will not be run, right?

andreaturli: replace `@BeforeMethod(alwaysRun = true)` with `@BeforeMethod`

kiuby88:
Squashing?

andreaturli:
@kiuby88 have you tested the uber-jar? does it work as the assembly version?

kiuby88:
Yes, I tested and it works fine

pom.xml
```diff
@@ -96,7 +96,7 @@
         <!-- Dependencies -->
         <guava.version>18.0</guava.version>
         <javatuples.version>1.2</javatuples.version>
-        <brooklyn.version>0.7.0-20150614.2158</brooklyn.version>
+        <brooklyn.version>0.8.0-SNAPSHOT</brooklyn.version>
```
kiuby88: Probably, I should change this vertion to 0.7.0-incubating, right?

andreaturli:
How different `deployer/src/test/resources/brooklyn-example-hello-world-webapp.war` from what inside plain brooklyn?

kiuby88:
Our .war uses env variables instead of java system properties. do you think it could be a good idea upload the brooklyn example in apache/icubator-brooklyn to use env variables if expected java system properties are null?

andreaturli:
don't know about that. let's leave it as-is for now.

Please squash the commits

pom.xml
andreaturli: I had already issued https://github.com/SeaCloudsEU/SeaCloudsPlatform/pull/88

kiuby88: :+1:

kiuby88:
could it be merged?

andreaturli:
thanks @kiuby88 ! Merging now!

kiuby88:
Thanks for the review @andreaturli , it was very useful!

PR #83, opened 2015-07-21
Pulling dependencies up to dependencyManagement on parent pom

mbarrientos:
Pulling up dependencies and also removing some duplicated declarations

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -92,17 +92,50 @@
 
         <includedTestGroups />
         <excludedTestGroups>Integration,Live</excludedTestGroups>
-        <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
+        
+        <!-- Dependencies -->
         <guava.version>18.0</guava.version>
         <javatuples.version>1.2</javatuples.version>
         <brooklyn.version>0.7.0-20150614.2158</brooklyn.version>
         <jetty.version>9.2.4.v20141103</jetty.version>
         <dropwizard.version>0.8.1</dropwizard.version>
+        <httpcomponents.version>4.3.6</httpcomponents.version>
+        <gson.version>2.3</gson.version>
+        <json.version>20140107</json.version>
+        <tomcat-servlet.version>8.0.21</tomcat-servlet.version>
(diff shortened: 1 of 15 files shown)


adriannieto:
@andreaturli could you review this?

pom.xml
```diff
@@ -92,17 +92,50 @@
...
+        <commons-httpclient.version>3.1</commons-httpclient.version>
+        <toscamodel.version>1.0.0</toscamodel.version>
+        <hsqldb.version>2.3.0</hsqldb.version>
+        <mysql-connector.version>5.1.27</mysql-connector.version>
+        <modaclouds.version>2.4</modaclouds.version>
+        <jaxb.version>2.2.6</jaxb.version>
+        <castor.version>1.2</castor.version>
+        
+        <!-- Test dependencies -->
         <junit.version>4.11</junit.version>
```
andreaturli: I think we should take the change to choose between junit or testng as having both already created some headaches.
I vote for testng as it supports groups more easily than junit, but not strong feelings.

No need to address this comment in this PR

pom.xml
```diff
@@ -331,27 +337,96 @@
                     <goals>deploy</goals>
                 </configuration>
             </plugin>
-            <plugin>
```
andreaturli: why do you want to remove this plugin?

sla/sla-core/pom.xml
```diff
@@ -144,13 +127,13 @@
 		<dependency>
 			<groupId>org.slf4j</groupId>
 			<artifactId>slf4j-simple</artifactId>
-			<version>${slf4j-version}</version>
```
andreaturli: good catch

pom.xml
mbarrientos: It was duplicated right above it (https://github.com/SeaCloudsEU/SeaCloudsPlatform/pull/83/files#diff-600376dffeb79835ede4a0b285078036R318)

andreaturli: right! missed that, thanks!

andreaturli:
+1

thanks @mbarrientos!

PR #85, opened 2015-07-25
feature/DISCOVERER (Final proto)

buccarel:
This is the final prototype of the Discoverer. It includes both the PaaSify and the CloudHarmony crawlers. Everything is accessible as RESTful service.
@szenzaro , since you are the one who uses the discoverer most, could you please review it?
Thanks.

### DIFF
--- a/discoverer/pom.xml
+++ b/discoverer/pom.xml
@@ -21,6 +21,7 @@
 		xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
 
 	<modelVersion>4.0.0</modelVersion>
+	<groupId>eu.seaclouds-project</groupId>
 	<artifactId>discoverer</artifactId>
 	<packaging>war</packaging>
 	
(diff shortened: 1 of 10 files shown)


discoverer/pom.xml
```diff
@@ -32,26 +33,41 @@
...
 			<groupId>com.googlecode.json-simple</groupId>
 			<artifactId>json-simple</artifactId>
+			<version>1.1.1</version>
 		</dependency>
+
+		<!-- test units -->
 		<dependency>
 			<groupId>org.testng</groupId>
 			<artifactId>testng</artifactId>
+			<version>${testng.version}</version>
```
szenzaro: you can remove the version inheriting it from the father

discoverer/pom.xml
```diff
@@ -32,26 +33,41 @@
...
+
+		<dependency>
+			<groupId>eu.seaclouds-project</groupId>
+			<artifactId>common</artifactId>
+			<version>0.8.0-SNAPSHOT</version>
+		</dependency>
+
+		<!-- JSON parser -->
 		<dependency>
 			<groupId>com.googlecode.json-simple</groupId>
```
szenzaro: this will probably be replaced by jackson.

szenzaro:
Some tests would be nice but overall it's a +1.

perezp:
Thank you for the PR, 

I agree with Simone that more test of the crawler should be provided.
Moreover, there are many defined Strings mixed the code, for example to define TOSCA yaml. 

It also is carrying some debt from the previous PR of the Discoverer. There is commented code and a folder "offerings_repo" is created in the user HOME. 

I merge it because we need to have this code inside and because it only modifies java files local to the Discoverer module, but we can undo the merge upon request.

PR #86, opened 2015-07-27
FEATURE(planner): new Matchmaker and Core

szenzaro:
@perezp can you review this PR?
Probably also @mbarrientos can have a look on this.


szenzaro:
I close this to update with the rest service

PR #87, opened 2015-07-27
FEATURE(planner): new core, matchmaker and service

szenzaro:
@mbarrientos, can you review this?


planner/core/src/main/java/eu/seaclouds/platform/planner/core/HttpHelper.java
```diff
@@ -0,0 +1,158 @@
...
+ * <p/>
+ * http://www.apache.org/licenses/LICENSE-2.0
+ * <p/>
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+public class HttpHelper {
```
mbarrientos: We should avoid creating more http clients and use java REST clients for the interaction between components. I guess it can be used temporarily, but IMHO it should eventually be replaced.

planner/core/src/main/java/eu/seaclouds/platform/planner/core/PlannerClient.java
```diff
@@ -0,0 +1,79 @@
...
+ * <p>
+ * http://www.apache.org/licenses/LICENSE-2.0
+ * <p>
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+public class PlannerClient {
```
mbarrientos: What is the purpose of this class? It looks like a test class

planner/matchmaker/pom.xml
```diff
@@ -30,8 +30,35 @@ limitations under the License.
     <dependencies>
         <dependency>
             <groupId>eu.seaclouds-project</groupId>
-            <artifactId>toscamodel</artifactId>
-            <version>1.0.0</version>
```
mbarrientos: I guess that the toscamodel repository can be now removed from planner pom, right?
(https://github.com/SeaCloudsEU/SeaCloudsPlatform/blob/master/planner/pom.xml#L50)

planner/matchmaker/pom.xml
```diff
@@ -30,8 +30,35 @@ limitations under the License.
...
+                <exclusion>
+                    <groupId>org.apache.brooklyn</groupId>
+                    <artifactId>brooklyn-software-webapp</artifactId>
+                </exclusion>
+            </exclusions>
+        </dependency>
+        <dependency>
+            <groupId>org.testng</groupId>
+            <artifactId>testng</artifactId>
+            <version>6.1.1</version>
```
mbarrientos: Please, remove the version tag to use the one declared at parent pom

planner/service/src/main/java/eu/seaclouds/platform/planner/service/DamGenResource.java
```diff
@@ -0,0 +1,63 @@
...
+    @GET
+    @Timed
+    public DamGeneratorResponse damgen(@QueryParam("adp") String adp){
+        return new DamGeneratorResponse(new Planner(genURLs, adp).generateDam());
+    }
+
+    @POST
+    @Timed
+    public DamGeneratorResponse damgenp(@QueryParam("adp") String adp){
+        return new DamGeneratorResponse(new Planner(genURLs, adp).generateDam());
```
mbarrientos: These methods are doing the same, but it seems that it's behaving more like a GET, right? Please remove the duplicated one.

planner/service/src/main/java/eu/seaclouds/platform/planner/service/PlanResource.java
```diff
@@ -0,0 +1,66 @@
...
+    @GET
+    @Timed
+    public PlannerResponse plan(@QueryParam("aam") String aam){
+        return getPlans(aam);
+    }
+
+    @POST
+    @Timed
+    public PlannerResponse planp(@QueryParam("aam") String aam){
+        return getPlans(aam);
```
mbarrientos: Same as before, these GET and POST methods are identical, please remove the duplicated POST

planner/service/src/main/java/eu/seaclouds/platform/planner/service/RePlanResource.java
```diff
@@ -0,0 +1,66 @@
...
+    @GET
+    @Timed
+    public PlannerResponse replan(@QueryParam("aam") String aam, @QueryParam("dam") String dam){
+       return getPlans(aam, dam);
+    }
+
+    @POST
+    @Timed
+    public PlannerResponse replanp(@QueryParam("aam") String aam, @QueryParam("dam") String dam){
+        return getPlans(aam, dam);
```
mbarrientos: Same as before, these GET and POST methods are identical, please remove the duplicated POST

planner/core/src/main/java/eu/seaclouds/platform/planner/core/HttpHelper.java
szenzaro: I agree, It should be something shared in common I think. I have seen that in the dashboard there is something similar. Do you agree?

planner/core/src/main/java/eu/seaclouds/platform/planner/core/PlannerClient.java
szenzaro: Oh. This was just for testing purposes. Ignore it.

planner/matchmaker/pom.xml
szenzaro: I left it there because of optimizer referencing to it. Otherwise CI complains. It should be removed when we have the updated optimized.

planner/matchmaker/pom.xml
szenzaro: Ok

planner/service/src/main/java/eu/seaclouds/platform/planner/service/DamGenResource.java
szenzaro: I added GET and POST of the same operation since I do not know which one will be used.
I think that we can just use POST. wdyt?

planner/service/src/main/java/eu/seaclouds/platform/planner/service/PlanResource.java
szenzaro: Same as before ;)

planner/service/src/main/java/eu/seaclouds/platform/planner/service/RePlanResource.java
szenzaro: Same as before ;)

planner/service/src/main/java/eu/seaclouds/platform/planner/service/DamGenResource.java
mbarrientos: If the service is retrieving information and not changing anything on the server it should stay as GET, but feel free to annotate it in the most accurate way.

szenzaro: Agreed on GET

planner/core/src/main/java/eu/seaclouds/platform/planner/core/PlannerClient.java
perezp: Great! Tests!! 
Could this be moved to src/tests/... in order to find it easier when looking for planner tests?

perezp:
There are some warnings while building the planner-service, finalizing with this. 

[WARNING] maven-shade-plugin has detected that some .class files
[WARNING] are present in two or more JARs. When this happens, only
[WARNING] one single version of the class is copied in the uberjar.
[WARNING] Usually this is not harmful and you can skeep these
[WARNING] warnings, otherwise try to manually exclude artifacts
[WARNING] based on mvn dependency:tree -Ddetail=true and the above
[WARNING] output

No problem if you were already aware of it. 

I don't have more comments. +1 .

planner/core/src/main/java/eu/seaclouds/platform/planner/core/PlannerClient.java
szenzaro: yes, even though this requires a running service to be performed.

andreaturli: Mark it as `Live` or `Integration` test then, please

szenzaro:
@perezp the jar warning depends on the fact that there are different dependencies that uses the same libraries but with different version. 
We should investigate more on that, but right now it does not compromise the functionalities.

szenzaro:
I have modified as you suggested. If everything is ok you can merge.

PR #88, opened 2015-07-28
bumped brooklyn version

andreaturli:


andreaturli:
good point @adriannieto! Updated!

usage/installer/src/main/assembly/files/blueprints/seaclouds-on-byon.yaml
```diff
@@ -19,12 +19,16 @@ services:
...
+    install.version: 0.7.0-incubating
     managementUser: admin
     managementPassword: p4ssw0rd
     brooklynLocalPropertiesContents: |
       brooklyn.webconsole.security.users=admin
       brooklyn.webconsole.security.user.admin.password=p4ssw0rd
+    brooklyn.config:
+      brooklynnode.download.archive.subpath: brooklyn-dist-0.7.0-incubating/
+      brooklynnode.classpath:
+        - classpath://deployer-0.7.0-incubating.jar
```
mbarrientos: I guess this should be `deployer-0.8.0-SNAPSHOT.jar`, right?

mbarrientos: Also, the `installer` module would need to include a dependency on `deployer` so that it's included on the classpath. However, if the purpose of this lines is including all the upcoming features of the `deployer` module (PR #82 ) it can be ommited in this one, I think.

andreaturli: Until we don't merge @kiuby88 PR it won't work either way, I think

andreaturli: I'm not sure the `installer`needs this dependency as it is needed at runtime by the deployer.

kiuby88: `deployer-${project.version}.jar` is added a resource to installer.jar, it allows to Brooklyn to find the deployer jar as a resource using the `classpath` notation. Then, as @mbarrientos mentioned, the deployer jar should be pointed as `classpath://deployer-0.8.0-SNAPSHOT.jar`, following the version of SeaCloudsPlatform.

andreaturli:
@kiuby88 I'll leave that with you, feel free to upgrade brooklyn version and test yaml files accordingly. Thanks

kiuby88:
:+1:

PR #92, opened 2015-08-07
fix for seaclouds-on-byon.yaml

andreaturli:


rosogon:
Tested it and sla and dashboard do not deploy. Are they working for you?

You must add MONITOR_METRICS_URL and SLA_URL back to sla entity.

andreaturli:
Thx @rosogon for the feedback. Can you please add whatever you think it is missing to this PR to have SLA working? I think you can either add another commit to my branch or open a new PR that supersedes this one or merge this and add a new PR to fix SLA. Wdyt?

rosogon:
@andreaturli : what is missing in the PR is in PR #78, that has already been accepted. 

I want to let it clear that the sla+dashboard not being deployed has nothing to do with #78.

If you add the content of #78, I can give you a +1 and work on the sla+dashboard fail in another PR. Wdyt?

rosogon:
Actually, the only missing var is SLA_URL, needed for the sla to attach as observer using its own URL as callback.

andreaturli:
@rosogon sorry but I didn't understand your comments. Can you rephrase them?

MicheleGuerriero:
@andreaturli @rosogon I had some tests with this PR in order to understand hat's going wrong. I report the results of my experiments.
I used brooklyn-0.7.0-incubating. I firt did some test locally and on EC2 deploying just the monitoring part isolated and the deploy succeed in both the cases. 
Deploying the entire platform the monitoring seems to be the only one successfully deployed. All the other components failed during my tests
The deployer  fails during the customize phase when brooklyn looks for this configuration 

brooklynnode.classpath:
        - classpath://deployer-0.8.0-SNAPSHOT.jar

 without find the .jar. I guess this .jar is the one I can also see in usage/compiler/target after compilation and I suppose this should be put somewhere inside the BROOKLYN_HOME in order to be visible. Do you think this could be the problem blocking the deployer? Where should the .jar file be put? I tried to put it in the lib/dropins and in the and lib/brooklyn folder of my brooklyn installation but it didn't work. 
Regarding the other components it seems (also simply looking at the yaml) that the dashboard is failing waiting for the deployer while the SLA deployment does not start and I cannot see any message about this fail (no activity is recorded by brooklyn about the SLA). The problem persists also adding @rogoson suggestions from PR #78. Any idea about why the deployment for the SLA is not starting?
I'll continue performing tests and trying to understand how to fix the deployer deployment and why the SLA deployment does not start, at least from my side. If you need any information, log, whatever from my experiments which could lead to better and faster diagnosis I can provide you with it.

andreaturli:
@MicheleGuerriero I've submitted a new version of the `seaclouds-on-byon.yaml` which works for me using this [notes](https://github.com/SeaCloudsEU/SeaCloudsPlatform/tree/master/usage/installer)

@SeaCloudsEU/developers If anyone can test it, I'll merge it as soon as I have a +1

mbarrientos:
Works fine for me on Ubuntu 14.04 machines. Thanks @andreaturli!

usage/installer/src/main/assembly/files/blueprints/seaclouds-on-byon.yaml
```diff
@@ -42,37 +51,24 @@ services:
...
-            MONITOR_METRICS_URL: >
-                $brooklyn:formatString("%s/v1/metrics", 
-                component("monitoring").component("child", "manager").attributeWhenReady("main.uri"))
-            SLA_URL: >
-                $brooklyn:formatString("http://%s:%s", 
-                component("sla-core").attributeWhenReady("host.address"), 
-                component("sla-core").attributeWhenReady("http.port"))
-      war: https://oss.sonatype.org/service/local/artifact/maven/redirect?r=snapshots&g=eu.seaclouds-project&a=sla-service&v=LATEST&e=war
-      install.latch: $brooklyn:component("sla-db").attributeWhenReady("service.isUp")
+      slaHost: $SLA_HOST
```
mbarrientos: I'm wondering why is this necessary? Is it any conflict with `attributeWhenReady` between `SameServerEntity` children?

andreaturli:
sorry @mbarrientos why are you using Ubuntu 14.04???
I think we want to use the Vagrantfile shipped with the projects which should install `precise pangolin` (12.04.x) so that anyone can start from the same environment.

I asked to use those notes [1 hour ago](https://github.com/SeaCloudsEU/SeaCloudsPlatform/pull/92#issuecomment-137389579)

sorry but it is really confusing

PR #94, opened 2015-09-11
Workaround to allow the wizard to skip the matchmake/optimization/dam-generation step.

adriannieto:
This will allow to the dashboard to skip the third step to be able to deploy applications with seaclouds if the user already has all the neccessary inputs (Deployment yaml, monitoring rules and sla agreements).

@andreaturli  could you take a look?


andreaturli:
lgtm, thanks @adriannieto

PR #96, opened 2015-09-17
Old monitor module content deleted and added data collectors for the NURO case study and to trigger reconfiguration

MicheleGuerriero:
This PR deletes the old and currently unused content of the monitor module and adds two new Tower 4Clouds Data Collectors providing new features. Each Data Collector is a single submodule of the main monitor module. The nuro-data-collector is basically a Tower 4Clouds customization for the NURO case study providing application specific metrics collected from the NURO sensors. The reconfiguration-data-collector right now collect information about the status of each application managed by SeaClouds by calling Brooklyn APIs. This DC periodically asks Brooklyn for the applications summary and for each application communicate to Tower 4Clouds the status and the hosting providers.


monitor/nuro-data-collector/pom.xml
```diff
@@ -0,0 +1,132 @@
...
+									<mainClass>dataCollector.Main</mainClass>
+								</manifest>
+							</archive>
+						</configuration>
+					</execution>
+				</executions>
+			</plugin>
+
+			<plugin>
+                <groupId>com.mycila</groupId>
```
andreaturli: this is not needed at this level, it may be a nice addition to the root pom.xml tho, please remove this plugin definition.

monitor/nuro-data-collector/src/main/java/config/Environment.java
```diff
@@ -0,0 +1,36 @@
...
+ *
+ *    Unless required by applicable law or agreed to in writing, software
+ *    distributed under the License is distributed on an "AS IS" BASIS,
+ *    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *    See the License for the specific language governing permissions and
+ *    limitations under the License.
+ */
+package config;
+
+public class Environment {
```
andreaturli: weird, why do you need this mapping?

monitor/nuro-data-collector/src/main/java/exception/MetricNotAvailableException.java
```diff
@@ -1,6 +1,6 @@
 /**
- * Copyright 2014 SeaClouds
- * Contact: Dionysis Athanasopoulos <dionysiscsuoi@gmail.com>
+ * Copyright 2015 SeaCloudsEU
+ * Contact: Michele Guerriero <michele.guerriero@mail.polimi.it>
```
andreaturli: do we really need that? not all the SeaClouds classes contain this header :(

monitor/nuro-data-collector/src/main/java/config/Environment.java
MicheleGuerriero: It was just to avoid to have strings within the code in the case in the future environmental variables would change (actually during twoer 4cloud developmente they have been changed several times even if now they should be stable). Anyway I can remove this mapping, do you suggest to hard code the variable names?

monitor/nuro-data-collector/src/main/java/exception/MetricNotAvailableException.java
MicheleGuerriero: I noticed that there were a kind of licensing process in SeaClouds and I though it was necessary to add this header. IN example if I'm not wrong the SLA module adopt the same way of licensing (by having an header file and a maven plugin setting this header on top of each class). Also other component classes are licensed, here's just an example https://github.com/SeaCloudsEU/SeaCloudsPlatform/blob/master/dashboard/src/main/java/eu/seaclouds/platform/dashboard/config/BasicEndpointFactory.java .Also here if this is wrong I can delete the header and the plugin. But maybe I just used the wrong ay of adding license?

monitor/nuro-data-collector/pom.xml
MicheleGuerriero: Ok I'm going to remove it.

monitor/nuro-data-collector/src/main/java/config/Environment.java
andreaturli: nope, just wondering why you need a class of constants: ;)

MicheleGuerriero: Oh ok, you mean to move those constants somewhere else. I will do this. Thanks for the suggestions!

PR #99, opened 2015-09-22
FEATURE(aamwriter): implemented aamwriter.

rosogon:
Generator of AAM from json topology) and REST endpoint.

This PR adds: 

* a module to planner (planner/aamwriter), where resides the functionality of the aamwriter;
* a REST endpoint to the dashboard (POST /api/aamwriter/translate)

Usage:

    cd planner/aamwriter
    java -jar target/aamwriter-jar-with-dependencies.jar <topology.json>

    curl -v -H"Content-type: application/json" -X POST -d@<topology.json> http://localhost:8000/api/aamwriter/translate

There are examples in src/main/resources/examples/.

### DIFF
--- a/dashboard/pom.xml
+++ b/dashboard/pom.xml
@@ -38,6 +38,12 @@
         </dependency>
         
         <dependency>
+            <groupId>eu.seaclouds-project</groupId>
+            <artifactId>aamwriter</artifactId>
+            <version>${project.version}</version>
+        </dependency>
+        
+        <dependency>
             <groupId>com.google.code.gson</groupId>
             <artifactId>gson</artifactId>
         </dependency>
--- a/dashboard/src/main/java/eu/seaclouds/platform/dashboard/DashboardApplication.java
+++ b/dashboard/src/main/java/eu/seaclouds/platform/dashboard/DashboardApplication.java
@@ -17,6 +17,7 @@
 
 package eu.seaclouds.platform.dashboard;
 
+import eu.seaclouds.platform.dashboard.resources.AamWriterResource;
 import eu.seaclouds.platform.dashboard.resources.DeployerResource;
 import eu.seaclouds.platform.dashboard.resources.MonitorResource;
 import eu.seaclouds.platform.dashboard.resources.PlannerResource;
@@ -56,5 +57,6 @@ public class DashboardApplication extends Application<DashboardConfiguration> {
         environment.jersey().register(new MonitorResource(configuration.getMonitorFactory()));
         environment.jersey().register(new PlannerResource(configuration.getPlannerFactory()));
         environment.jersey().register(new SlaResource(configuration.getSlaFactory()));
+        environment.jersey().register(new AamWriterResource());
     }
 }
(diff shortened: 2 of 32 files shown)


andreaturli:
@rosogon I've a general comment: is it possible to generate a YAML representation from the GUI, without generating json first that need to be translated then?

rosogon:
The json file is actually the tree of objects that represent the topology in memory, and is used by the d3 and the code that handles the forms. So, in any case, a translation to TOSCA must be implemented, either in javascript in the client or in the server.

planner/aamwriter/src/main/resources/examples/drafts/simple_adp.yaml
```diff
@@ -0,0 +1,104 @@
...
+            properties:
+                username: dbuser
+                password: dbpassword
+            requirements:
+                - host: pivotal
+                
+        aws-ec2:us-west-2:
+            type: seaclouds.nodes.Compute.Amazon
+            properties:
+              type: compute
```
szenzaro: This should be "resource_type"

planner/aamwriter/src/main/resources/examples/drafts/simple_adp.yaml
```diff
@@ -0,0 +1,104 @@
...
+              num_cpus: 4
+              availability: 0.98
+              cost: 0.928 usd_per_hour
+              performance: 62 ecb
+    
+        pivotal:
+            type: seaclouds.nodes.Platform.Pivotal
+            properties:
+              type: platform
```
szenzaro: This should be "resource_type"

planner/aamwriter/src/main/resources/examples/drafts/simple_adp.yaml
```diff
@@ -0,0 +1,104 @@
...
+              mysql_support: true
+              postgresql_support: true
+              mongoDB_support: true
+              redis_support: true
+              riak_support: true
+              dataStax_support: true
+              neo4j_support: true
+              pivotalHD_support: true
+              price: 0.03 $_perGB_per_h
+
```
szenzaro: We must add the java_version property "java_version: 7" for example

planner/aamwriter/src/main/resources/examples/drafts/simple_dam.yaml
```diff
@@ -0,0 +1,48 @@
+
```
szenzaro: No need to add this, because the DAM looks differently, I can provide a full DAM example if needed.

szenzaro:
One last comment is about squashing the two commits into a single one.
After that I think we can merge

planner/aamwriter/src/main/resources/examples/drafts/simple_adp.yaml
rosogon: All the files in examples/drafts are drafts and should not taken into account. I will remove them.

andreaturli:
@rosogon I was wondering why haven't you used the a4c classes like NodeType or ArchiveRoot instead of implementing your own? 

I think we spent a lot of effort for the ToscaParser and we should use it as much as possible: that's why it is in common module, after all.

Thoughts?

Thanks

rosogon:
Simone suggested me to build a snake yaml object and serialise that ("Unfortunately with a4c only the topology is serializable unless you want to extend their serializer"). And we are using in the AAM node_types and groups.

andreaturli:
@rosogon do you think we could use this library http://jsontoyaml.com/#javascript instead?
Looks like it is able to transform json-formatted objects into yaml-formatted objects quite easily, so that we can avoid the java rest endpoint ?

PR #101, opened 2015-09-23
Third version of optimizer

perezp:
This version of the Optimizer is able to accept inputs in the TOSCA syntax defined during the SeaClouds General Assembly in September 2015. 
The functionality of the module has not been increased. It is the same as in the previous versions. 
Bugs that have been found in the mean time have been corrected

@adriannieto  please, could you review it?


perezp:
I did the commit squash since the last PR of the optimizer (April). I still see here the commits from February to April... I'm solving it. @adriannieto

adriannieto:
As you already said, we cannot review 100 commits :smile: try to find a way to solve this, maybe it's enough to do a reset soft + push stash + reset to master branch + pop stash + commit.

kiuby88:
It could be useful for you
http://git-scm.com/docs/git-cherry-pick

PR #102, opened 2015-09-24
integrate planner yaml

andreaturli:
- add planner to seaclouds-on-byon.yaml
- remove authN from deployer
- fix seaclouds.yaml blueprint (deployed on AWS EC2)


szenzaro:
+1 for me

PR #103, opened 2015-09-25
FIX: solving a services management issue.

kiuby88:
The application tried to bind a non-existent service.


kiuby88:
@mbarrientos, could you take a look please?

PR #104, opened 2015-09-29
(Feature) Planner and discoverer blueprints for Brooklyn

szenzaro:
Added blueprints and readme for planner and discoverer

### DIFF
--- a/usage/installer/src/main/resources/eu.seaclouds.planner/planner_blueprint.yml
+++ b/usage/installer/src/main/resources/eu.seaclouds.planner/planner_blueprint.yml
@@ -1,26 +1,22 @@
-name: SeaClouds Planner Rest Service
-location: localhost
+name: SeaClouds Planner
 services:
 - type: brooklyn.entity.basic.VanillaSoftwareProcess
-  name: Rest Service
-
+  name: Planner Rest Service
   brooklyn.config:
-    planner.install.version: "0.8.0-20150911.114153-8"
     planner.config.port: 1234
     planner.config.discoverer.url: undefined
     planner.config.optimizer.url: undefined
-    planner.config.credential_generator.url: undefined
     planner.config.monitor_generator.url: undefined
     planner.config.sla_generator.url: undefined
+    planner.config.service.url: https://oss.sonatype.org/service/local/artifact/maven/redirect?r=snapshots&g=eu.seaclouds-project&a=planner-service&v=LATEST&e=jar
 
   shell.env:
-    VERSION: $brooklyn:config("planner.install.version")
     PLANNER_PORT: $brooklyn:config("planner.config.port")
     DISCOVERER_URL: $brooklyn:config("planner.config.discoverer.url")
     OPTIMIZER_URL: $brooklyn:config("planner.config.optimizer.url")
-    CREDENTIAL_GENERATOR_URL: $brooklyn:config("planner.config.credential_generator.url")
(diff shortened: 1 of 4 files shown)


andreaturli:
Thx @szenzaro 
Given these updates, do we need to update also 'seaclouds-on-byon' and 'seaclouds' blueprints?

szenzaro:
@andreaturli, yes I updated them. Could you please check their correctness?

I detached the planner as a child from the Dashboard since the planner is an independent service.

usage/installer/src/main/resources/eu.seaclouds.planner/planner_blueprint.yml
```diff
@@ -0,0 +1,39 @@
...
+
+  shell.env:
+    PLANNER_PORT: $brooklyn:config("planner.config.port")
+    DISCOVERER_URL: $brooklyn:config("planner.config.discoverer.url")
+    OPTIMIZER_URL: $brooklyn:config("planner.config.optimizer.url")
+    MONITOR_GENERATOR_URL: $brooklyn:config("planner.config.monitor_generator.url")
+    SLA_GENERATOR_URL: $brooklyn:config("planner.config.sla_generator.url")
+    PLANNER_SERVICE_URL: $brooklyn:config("planner.config.service.url")
+
+  launch.command: |
```
rosogon: Is not there a simpler way for executing a jar?

szenzaro: @andreaturli helped me to write this. If there is a simpler way to write it he would probably know better than me. Andrea, should we change this one?

rosogon: If Andrea helped you, then it should be the way. So, no need to change. I simply thought it was going to be easier.

rosogon:
@szenzaro . This branch has conflicts. You should fix them before it can be merged.

usage/installer/src/main/resources/eu.seaclouds.planner/planner_blueprint.yml
szenzaro: I am not an expert of Brooklyn blueprints, so I really do not know if there is a better way to write it.

rosogon:
seaclouds-on-byon gives the following error:  Invalid YAML: null; mapping values are not allowed here; in 'reader', line 79, column 11: name: Discoverer ^

szenzaro:
I removed the seaclouds on byon and seaclouds.yaml.

rosogon:
@szenzaro : Can we close this PR?

szenzaro:
I think this one should be merged after the #105. If there are other reasons why we should close this, let me know.

rosogon:
Hi @szenzaro . #109 changes location of blueprints, and already has the blueprint for the planner. Maybe @andreaturli can say if this PR is needed or not.

andreaturli:
if we merge https://github.com/SeaCloudsEU/SeaCloudsPlatform/pull/109 (which is *highly* recommended!!) we won't need these single yaml files anymore.

szenzaro:
Ok so we can close this PR. But we need to add the discoverer blueprint as well.

andreaturli:
please see https://github.com/SeaCloudsEU/SeaCloudsPlatform/pull/109, it now contains also the discoverer. Tested already on BYON

PR #105, opened 2015-09-30
(Feature) New planner version

szenzaro:
Updated planner version.


rosogon:
Hi @szenzaro . Could you summarize the status of this planner version with respect to the integration with other components? (optimizer, monitoring rules generator, sla generator).

szenzaro:
The new version includes, complete matchmaker and DAM generator, and the initial support for replanning features.
In particular the planner reads the aamwriter output as input for planning and understand the intended output of monitor generator, and the agreed sla generator output.

About the integration with the corresponding services, since some of them are not ready and the optimizer is in PR state, I used sample outputs from each component when available and manually generated examples of agreed interfaces when not available.

With dummy services it works, we must test the integration as soon as the other services are ready.

The integration with the discoverer has been successfully achieved.

rosogon:
@szenzaro : I tried to test the REST interface. What is the right way to call it?

```
$ curl localhost:1234/plan -H"Accept:application/json"
<html>
<head>
<meta http-equiv="Content-Type" content="text/html; charset=UTF-8"/>
<title>Error 404 Not Found</title>
</head>
<body><h2>HTTP ERROR 404</h2>
<p>Problem accessing /plan. Reason:
<pre>    Not Found</pre></p><hr><i><small>Powered by Jetty://</small></i><hr/>

</body>
</html>
```

rosogon:
I noticed the REST interface receives the AAM as a query param. You should receive it in the body, don't you think?

szenzaro:
@rosogon after running the application via "java -jar servicename.jar server plannerconf.yml"

you will find the api at host:1234/planner/ ; so to plan: host:1234/planner/plan

It is a query param because it is implemented as a GET instead of a POST. If I remember well, I changed it to GET from POST after talking with @adriannieto

adriannieto:
About GET/POST i just did further reading and I'm not sure, GET its a good idea because it doesn't modify any status in the server, but POST allows us to send data inside the body. 

I'm confused about how to proceed, what's the best solution? (Get + body breaks the http spec as far as I know)

rosogon:
For the record. To manually call the API: 

curl -v http://localhost:1234/planner/plan --data-urlencode aam@"planner/aamwriter/src/main/resources/examples/web-chat_aam.yml" -G -X GET

rosogon:
@szenzaro: I started discoverer and planner. 

discovererURL=http://127.0.0.1:1235/

It looks it cannot connect to discoverer. Exception:
```
org.apache.http.client.ClientProtocolException
        at org.apache.http.impl.client.InternalHttpClient.doExecute(InternalHttpClient.java:186)
        at org.apache.http.impl.client.CloseableHttpClient.execute(CloseableHttpClient.java:82)
        at org.apache.http.impl.client.CloseableHttpClient.execute(CloseableHttpClient.java:106)
        at eu.seaclouds.platform.planner.core.HttpHelper.getRequest(HttpHelper.java:68)
        at eu.seaclouds.platform.planner.core.Planner.getOfferingsFromDiscoverer(Planner.java:267)
        at eu.seaclouds.platform.planner.core.Planner.plan(Planner.java:111)
        ...
Caused by: org.apache.http.ProtocolException: Target host is not specified
        at org.apache.http.impl.conn.DefaultRoutePlanner.determineRoute(DefaultRoutePlanner.java:69)
        at org.apache.http.impl.client.InternalHttpClient.determineRoute(InternalHttpClient.java:124)
        at org.apache.http.impl.client.InternalHttpClient.doExecute(InternalHttpClient.java:183)
        ... 63 more
```

szenzaro:
I will take a look, thanks.

adriannieto:
I was reading further and I try feel that POST+body is maybe the best way, for example aamwriter uses it.

It shouldn't be a big deal to change it, so consider it as a suggestion.

Anyone do anyone have another opinion?

rosogon:
Hi, @szenzaro. Could you squash in one commit one and rebase? I'll merge it once done.

rosogon:
@adriannieto : The problem is that generate the aam or the plan are commands, not operations on resources. We can RESTfully this inventing a command resource :)

szenzaro:
@rosogon I have squashed the commits.

andreaturli:
@SeaCloudsEU/developers Anyone tested this PR before merging it?

The deployment is broken again!

```
$ java -jar planner-service.jar server conf.yml
conf.yml has an error:
  * Unrecognized field at: credentialGeneratorURL
    Did you mean?:
      - slaGeneratorURL
      - monitorGeneratorURL
      - discovererURL
      - optimizerURL
      - server
        [2 more]
```

@szenzaro, any ideas?

szenzaro:
See #117 .

PR #106, opened 2015-10-01
FEATURE: Third version of optimizer. Input/output format

perezp:
This version of the Optimizer is able to accept inputs in the TOSCA syntax defined during the SeaClouds General Assembly in September 2015. 
The functionality of the module has not been increased. It is the same as in the previous versions. 
Bugs that have been found in the mean time have been corrected

@adriannieto please, could you review it?
Hopefully the issues with hundreds of commits is solved

### DIFF
--- a/planner/optimizer/src/main/java/eu/seaclouds/platform/planner/optimizer/Optimizer.java
+++ b/planner/optimizer/src/main/java/eu/seaclouds/platform/planner/optimizer/Optimizer.java
@@ -24,8 +24,7 @@ import eu.seaclouds.platform.planner.optimizer.heuristics.SearchMethodName;
 
 public class Optimizer {
 
-   private static Logger          log = LoggerFactory
-                                            .getLogger(Optimizer.class);
+   private static Logger          log = LoggerFactory.getLogger(Optimizer.class);
    private final int              NUMBER_OF_PLANS_GENERATED;
    private final SearchMethodName searchName;
 
@@ -58,29 +57,26 @@ public class Optimizer {
       outputPlans[0] = "Plan generation was not possible";
 
       if (previousPlans == null) {
-         OptimizerInitialDeployment initialOptimizer = new OptimizerInitialDeployment(
-               searchName);
+         OptimizerInitialDeployment initialOptimizer = new OptimizerInitialDeployment(searchName);
 
          try {
-            outputPlans = initialOptimizer.optimize(appModel,
-                  suitableCloudOffer, NUMBER_OF_PLANS_GENERATED);
+            outputPlans = initialOptimizer.optimize(appModel, suitableCloudOffer, NUMBER_OF_PLANS_GENERATED);
             previousPlans = outputPlans;
          } catch (Error E) {
             log.error("Error optimizing the initial deployment");
-
          }
       } else {
          Reoptimizer optimizerReplanning = new Reoptimizer(searchName);
 
-         try {
+         try{
             log.error("Calling a Replanning. The previously generated Plan will be used as a base");
-            outputPlans = optimizerReplanning.optimize(appModel,
-                  suitableCloudOffer, NUMBER_OF_PLANS_GENERATED);
+            outputPlans = optimizerReplanning.optimize(appModel, suitableCloudOffer, NUMBER_OF_PLANS_GENERATED);
             previousPlans = outputPlans;
          } catch (Error E) {
             log.error("Error optimizing the Replanning");
-
+            throw E;
          }
+
       }
       return outputPlans;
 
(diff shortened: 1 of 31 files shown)


planner/optimizer/src/main/java/eu/seaclouds/platform/planner/optimizer/Optimizer.java
```diff
@@ -74,8 +71,7 @@ public Optimizer(int num, SearchMethodName name) {
 
          try {
             log.error("Calling a Replanning. The previously generated Plan will be used as a base");
-            outputPlans = optimizerReplanning.optimize(appModel,
-                  suitableCloudOffer, NUMBER_OF_PLANS_GENERATED);
+            outputPlans = optimizerReplanning.optimize(appModel, suitableCloudOffer, NUMBER_OF_PLANS_GENERATED);
             previousPlans = outputPlans;
          } catch (Error E) {
             log.error("Error optimizing the Replanning");
```
rosogon: If an exception occurs without taking any corrective action, you should bubble up the exception. If not, the error is masked. I would enclose the Error inside a OptimizerException, which extends RuntimeException.

planner/optimizer/src/main/java/eu/seaclouds/platform/planner/optimizer/OptimizerInitialDeployment.java
```diff
@@ -42,69 +49,222 @@ public OptimizerInitialDeployment() {
...
-         log.error("Topology could not be parsed. Not known quantity of calls between modules. Assuming the dummy case where"
-               + "all modules are called in sequence. The order of calls is random}");
+         log.error(
+               "Topology could not be parsed. Not known quantity of calls between modules. Assuming the dummy case where"
+                     + "all modules are called in sequence. The order of calls is random}");
          topology = createAdHocTopologyFromSuitableOptions(appInfoSuitableOptions);
       }
 
+      // Read the quality requirements of the application
+      if (IS_DEBUG) {
```
rosogon: If this is debugging info, you should simply log.debug

planner/optimizer/src/main/java/eu/seaclouds/platform/planner/optimizer/OptimizerInitialDeployment.java
```diff
@@ -42,69 +49,222 @@ public OptimizerInitialDeployment() {
...
       }
 
       return stringSolutions;
    }
 
+   private Map<String, Object>[] hashMapOfFoundSolutionsWithThresholds(Solution[] bestSols,
+         Map<String, Object> applicMap, Topology topology, SuitableOptions cloudOffers, int numPlansToGenerate,
+         QualityInformation requirements) {
+
+      if (IS_DEBUG) {
```
rosogon: You can use log.isDebugEnabled().

perezp: Thanks. I have used it a lot of times along. I'll change it. 
In the beggining it was with log.debug, but at some point this log.debug stopped working and I could not change the log level to show the debug messages (when we changed to testng). I did not look for a solution in that moment. I'll do it now.

rosogon:
@perezp : Ufff... The PR is too big. General comments:

* IMO, it's legal to log.debug or not depending in a variable in your code if you need finer grain. But use log.debug, and not log.info.
* Do not comment large portions of code to have them just in case. That's what the repository is for.
* Java coding standards state that methods name start with lowercase letter.

Is the class Optimizer the "main" class?

Sorry, for not reviewing this before. I had not realized you already resubmitted.

adriannieto:
@perezp sorry to not realizing about this until now. I agree with @rosogon suggestions. I will continue the review with line comments.

planner/optimizer/src/main/java/eu/seaclouds/platform/planner/optimizer/Optimizer.java
perezp: Right. Done.
I've modified it (not yet pushed). Now the the same error is thrown in the replanning case (infact it is not implemented yet the replannig). 

In case of the initial optimizer (line 66), I left the error still masked. I think this is still OK, as the this optimize method is the main one called (as a service in the near future), masking the error but returning empty plans I think is a good decision at this moment. 

Let me know if you disagree on this.

perezp:
@rosogon Regarding general comments. 

- "IMO, it's legal to log.debug or not depending in a variable in your code if you need finer grain. But use log.debug, and not log.info. "
Right. DONE, as you suggested. I liked to use the variable because during debugging I wanted to make a class very verbose, but not the rest of classes. I could not change the log level after the first instantiation.  

- "Do not comment large portions of code to have them just in case. That's what the repository is for."
You are right. I was my fault that I forgot to remove these methods. 

- "Java coding standards state that methods name start with lowercase letter."
You're right. DONE.

"Is the class Optimizer the "main" class?"
Yes, in the next PR , which hopefully will be just a couple of days after the acceptance of this one, the Optimizer class will be a web service.

Sorry, for not reviewing this before. I had not realized you already resubmitted.
No problem. Thanks.

rosogon:
Merged. @perezp, please provide REST service asap.

PR #110, opened 2015-10-06
Added Brooklyn entities for PHP applications.

mbarrientos:
Updated support for PHP applications on Brooklyn. This is based on the initial work done by @kiuby88 on the fork of incubator-brooklyn at SeaCloudsEU.

Apache2 is being configurated by default on machines where the apps are deployed. Apps can be deployed either by providing a tarball or a valid git repository.

### DIFF
--- a/deployer/pom.xml
+++ b/deployer/pom.xml
@@ -55,6 +55,16 @@
         </dependency>
 
         <dependency>
+            <groupId>org.apache.brooklyn</groupId>
+            <artifactId>brooklyn-software-webapp</artifactId>
+        </dependency>
+        
+        <dependency>
+            <groupId>org.apache.brooklyn</groupId>
+            <artifactId>brooklyn-software-database</artifactId>
+        </dependency>
+        
+        <dependency>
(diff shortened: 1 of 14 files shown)


andreaturli:
Thanks @mbarrientos and @kiuby88 

It is a great start but it would be nicer if Apache HTTPD entity were simpler.

I think we could do the following:
- have only 1 entity: HTTPD
- add a configKey for apache mod, this configKey is a Map<String, List<String>>

As an example, you can pass something like: <"libapache2-mod-php5", [ "php5" ] >
that are the name of the module, libapache2-mod-php5", and a list of dependencies for that module, [ "php5" ], using internally `BashCommands.installPackage()` to install those deps.

wdyt?

andreaturli:
Although the code needs to be simplified, we can address my general comment on a subsequent PR.

Merging

kiuby88:
Thx @andreaturli

kiuby88:
@andreaturli  I think it is a good idea for the following steps :+1:  I will need your support for creating this entity in the correct way.

PR #111, opened 2015-10-07
Added a Api method to retrieve the location of all the SeaClouds Components

adriannieto:
@MicheleGuerriero  could you review this? This adds also the necessary method to retrieve where Grafana is deployed, so after the installation script become fixed it the dashboard will be wired properly.

### DIFF
--- a/dashboard/config.yml
+++ b/dashboard/config.yml
@@ -16,6 +16,10 @@ deployer:
 monitor:
   host: ${MONITOR_HOST}
   port: ${MONITOR_PORT}
+  graphiteHost: ${GRAPHITE_HOST}
+  graphitePort: ${GRAPHITE_PORT}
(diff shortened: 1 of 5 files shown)


andreaturli:
@adriannieto this branch seems to have conflicts. Can you check it, please?

adriannieto:
@andreaturli Since the installation method was upgrade, I'm checking if everything still works.

adriannieto:
@andreaturli working now. The catalog currently has a fake graphite/graphana endpoint. I will have the proper one when @MicheleGuerriero  will find a fix for Graphite &  Grafana installation

andreaturli:
@adriannieto lgtm

Could you please confirm that the seaclouds deployment is still working after this PR got merged?

PR #115, opened 2015-10-08
First version of the monitoring-dam-generator

MicheleGuerriero:
This PR adds a first version of the monitoring-dam-generator. This component is called by the DAM generator in order to obtain deployment related monitoring information such as the monitoring rules to be installed into Tower 4Clouds and the data collectors that need to be deployed along with scripts for doing this. The monitoring-dam-generator takes as input the ADP. In particular for each application module found the monitoring-dam-generator is able to produce the following information:

-  a set of monitoring rules required for the metrics visualization, in particular for the module response time, if the module is a java application and, if the module is deployed on IaaS, for the hosting machine cluster average CPU and Ram utilization.

- a set of monitoring rules required for the SLA process. In this first version only the application availability and response time, if the module is a java application, are considered for possible SLAs.

- a monitoring rule required for the replanning process. In particular this monitoring rule looks for the isOnFireApp metric, which help to check the status of the application, collected by the reconfiguration-data-collector.

- if the application is the NURO case study a set of specific monitoring rules is produced for collecting metric from the NURO sensors via the nuro-data-collector.

The monitoring-dam-generator also generate deployment bash scripts for each of the 4 currently managed data collectors (nuro-data-collector, reconfiguration-data-collector, java-app-dc, modaclouds-dc).
Lunching the jar it expose a Rest service listening by default on port 8176, which can be called invoking a POST on the path /data  (i.e. POST 127.0.0.1:8176/data). The body of the request must be a plain text containing the yaml of the ADP.


andreaturli:
@MicheleGuerriero do we really need to add another rest server? Is there any components that needs to call the dam-generator from remote? The architectural overhead introduced by the rest server doesn't seem to be justified. 

More comments on the code soon.

dinitto:
@andreaturli: I agree that, on the one hand, there will be an overhead. On the other hand, building this as a separate service will increase its independency from the other parts and therefore its exploitability as a separate component. We have extensively discussed this point in previous meetings and decided to privilege this approach over tighter connections between components.

MicheleGuerriero:
I understood from previous telcos and also talking with @szenzaro that we need a REST service for this component, that's why I provided it. Anyway I think it should be also possible to simply import as a dependency the component and to call the method receiveData(InputStream adp) directly instantiating an object of type MonitoringDamGenerator, if we decide to work in this way.

szenzaro:
We do not really "need" a rest service for every component. As @dinitto said, it has been decided to adopt the REST microservices approach, and I followed this decision.

monitor/monitoring-dam-generator/src/test/java/eu/seaclouds/monitor/monitoringDamGenerator/MainTest.java
```diff
@@ -0,0 +1,193 @@
...
+                                        .newInstance("it.polimi.tower4clouds.rules");
+                                Unmarshaller unmarshaller = context.createUnmarshaller();
+                                InputStream stream = new ByteArrayInputStream(policy.get(key).toString().getBytes());
+                                MonitoringRules rules=(MonitoringRules) unmarshaller.unmarshal(stream);
+                                for(MonitoringRule rule: rules.getMonitoringRules()){
+                                    String ruleId=rule.getId();
+                                    switch(ruleId){
+                                   
+                                    case "respTimeRule_"+test_module_name: 
+                                        Assert.assertEquals(rule.getTimeStep(), "10");
```
andreaturli: create a method to assert everything so that you can reuse it later

monitor/monitoring-dam-generator/src/main/java/eu/seaclouds/monitor/monitoringDamGenerator/util/RuleSchemaGenerator.java
```diff
@@ -0,0 +1,150 @@
...
+        monitoredTarget = factory.createMonitoredTarget();
+
+        if (targetClass == null) {
+            monitoredTarget.setClazz("InternalComponent");
+        } else {
+            monitoredTarget.setClazz(targetClass);
+        }
+
+        if (targetType == null) {
+            logger.error("A valid target type must be specified", new RuntimeException("A valid target type must be specified"));
```
andreaturli: this will log, but not exit from the method, is it intended?

MicheleGuerriero: No, that's my fault. I will correct this.

monitor/monitoring-dam-generator/src/test/java/eu/seaclouds/monitor/monitoringDamGenerator/MainTest.java
MicheleGuerriero: Ok, right, will be done.

PR #118, opened 2015-10-14
FEATURE(sla): Added endpoint to create a SLA Template from the DAM.

rosogon:
Added SLA Generation from the DAM. This is needed in the DAM Generation process. This commit adds:

* Generation of template from DAM 
* Partial generation of agreement from template
* REST endpoint to add template from DAM
* REST endpoint to add agreement from template (TBF)

To test:
* run sla service: `java -jar jetty-runner.jar --path / sla-service/target/sla-service.war`
* post DAM: `curl http://localhost:8080/seaclouds/templates -X POST --data-binary @"sla-generator/src/test/resources/DemoDAM.yml"`


sla/sla-core/pom.xml
```diff
@@ -113,17 +114,17 @@
 
 	<dependencies>
 		<dependency>
-            <groupId>junit</groupId>
-            <artifactId>junit</artifactId>
-            <version>${junit.version}</version>
-            <scope>test</scope>
+			<groupId>junit</groupId>
```
andreaturli: why do you need both junit and testng ?

sla/sla-core/pom.xml
```diff
@@ -113,17 +114,17 @@
...
+			<artifactId>junit</artifactId>
+			<version>${junit.version}</version>
+			<scope>test</scope>
+		</dependency>
+		<dependency>
+			<groupId>org.testng</groupId>
+			<artifactId>testng</artifactId>
+			<version>${testng.version}</version>
```
andreaturli: remove version as inherited from the parent

sla/sla-core/pom.xml
```diff
@@ -113,17 +114,17 @@
...
+			<scope>test</scope>
 		</dependency>
-        <dependency>
-            <groupId>org.testng</groupId>
-            <artifactId>testng</artifactId>
-            <version>${testng.version}</version>
-            <scope>test</scope>
-        </dependency>
 		<dependency>
```
andreaturli: we use `log4j12`in SeaClouds instead of `slf4j-simple`. Unless strictly needed, prefer log4j

sla/sla-core/sla-generator/src/test/java/eu/seaclouds/platform/sla/generator/SlaGeneratorTest.java
```diff
@@ -0,0 +1,117 @@
...
+import eu.seaclouds.platform.sla.generator.AgreementGenerator;
+import eu.seaclouds.platform.sla.generator.ContextInfo;
+import eu.seaclouds.platform.sla.generator.JaxbUtils;
+import eu.seaclouds.platform.sla.generator.SlaInfo;
+import eu.seaclouds.platform.sla.generator.TemplateGenerator;
+import eu.seaclouds.platform.sla.generator.ContextInfo.Validity;
+
+public class SlaGeneratorTest {
+
+    @BeforeMethod
```
andreaturli: why this empty @BeforeMethod ?

sla/sla-core/sla-generator/src/test/java/eu/seaclouds/platform/sla/generator/SlaGeneratorTest.java
```diff
@@ -0,0 +1,117 @@
...
+    public void beforeMethod() {
+    }
+
+    @BeforeClass
```
andreaturli: why this empty @BeforeClass ?

sla/sla-core/pom.xml
rosogon: The TestNG eclipse plugin converts JUnit tests to TestNG tests, but the ones run with SpringRunner. The process for these tests is not easy and has problems (http://stackoverflow.com/questions/2608528/spring-dependency-injection-with-testng, http://stackoverflow.com/questions/5192562/spring-autowiring-happens-after-beforeclass-when-running-test-with-maven-surefi).

Having both dependencies, the effect is that TestNG is able to run the JUnit tests (http://testng.org/doc/migrating.html).

sla/sla-core/pom.xml
rosogon: This is only in test scope, to see logs on stdout. I removed it, anyway.

rosogon:
@andreaturli, could you finish the review? Thanks

andreaturli:
sorry, don't have time to finish that. Could you suggest another reviewer?

sla/sla-core/sla-personalization/pom.xml
```diff
@@ -53,6 +59,11 @@
 			<groupId>com.sun.jersey</groupId>
 			<artifactId>jersey-client</artifactId>
 		</dependency>
+
+		<dependency>
+			<groupId>it.polimi.modaclouds.qos-models</groupId>
+			<artifactId>qos-models</artifactId>
+		</dependency>
```
MicheleGuerriero: As you know this library is deprecated, we should move to the **rules** module of Tower 4Clouds, version 0.2.3 ( https://github.com/deib-polimi/tower4clouds/releases/download/v0.2.3/rules-0.2.3-jar-with-dependencies.jar ). We already discussed this in pvt and maybe the best solution is to put this T4C dependency (along with the repository) in the root pom, since it is shared by the monitor modules and SLA. Anyway right now I'm placing this dependency in the modules' pom, where required, can you do the same here at least for now?

MicheleGuerriero:
Hi @rosogon and @andreaturli , I reviewed this PR since I was previously assigned to it. Just a comment on the dependency from Tower 4Clouds that maybe can be updated in this PR. The rest of the code looks fine at least to me.

sla/sla-core/sla-personalization/pom.xml
rosogon: Of course. Just paste here the needed dependency. When both PRs have been accepted, we can submit one with the dependency in the root.

MicheleGuerriero: Here is the one I'm currently using in the monitoring-dam-generator 

        <dependency>
        <groupId>it.polimi.tower4clouds</groupId>
        <artifactId>rules</artifactId>
        <version>0.2.3</version>
    </dependency>    

Of course you need also the repository declaration.

rosogon:
@MicheleGuerriero : is it ok now?

MicheleGuerriero:
+1

PR #119, opened 2015-10-14
SQUASH: squasing last commits for "Optimizer service" task

perezp:
This PR exposes the Seaclouds optimizer as a service. 
Optimizer functionality has not changed at all in this period. 
This  PR includes tests of the service running on Jetty. 
Although there are changes in 12 files, significant changes are only in 3 of those files, the ones with the larger quantity of modifications.

In the next PR I'll fill the part of the optimizer in the file .bom for brooklyn for its deployment with the platform.


andreaturli:
I think #122 supersedes this one, so it can be closed

PR #120, opened 2015-10-15
Unused monitor content deleted.

MicheleGuerriero:
This PR simply removes the old unused monitor modules and clean the monitor module's pom. This is in preparation of the next PRs that will add each new module step by step. In particular having the pom already clean will prevent the next PRs from having conflicts between each other.


andreaturli:
thanks @MicheleGuerriero 

Merging now :+1:

PR #121, opened 2015-10-15
Fixed initialization of the dashboard inside the catalog

adriannieto:
A typo in the catalog was causing the Dashboard to fail while looking for the required Components. @andreaturli could you take a look?


andreaturli:
thanks @adriannieto

PR #122, opened 2015-10-15
Optimizer both versions

perezp:
This PR creates the Seaclouds optimizer as a jar (optimizer-core) for importing and java method invocation, and also as a service with war packaging (optimizer-service).
Optimizer functionality has not changed at all in this period. 
This PR includes tests of the service running on Jetty to crate the war.
All files have changed, mainly due to the refactoring.

This PR is in conflict with the previous one of the optimizer. 
@andreaturli @adriannieto @rosogon @szenzaro @MicheleGuerriero @kiuby88 and everybody who may be interested in today's discussion during the integration telco. If any of you are happy to review the changes in this PR I discard the previous PR.

In the next PR I'll fill the part of the optimizer in the file .bom for brooklyn for its deployment with the platform.


planner/optimizer/optimizer-core/pom.xml
```diff
@@ -0,0 +1,31 @@
...
+    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
+  <modelVersion>4.0.0</modelVersion>
+  <parent>
+    <groupId>eu.seaclouds-project</groupId>
+    <artifactId>optimizer</artifactId>
+    <version>0.8.0-SNAPSHOT</version>
+  </parent>
+  <groupId>eu.seaclouds-project</groupId>
+  <artifactId>optimizer-core</artifactId>
+  <version>0.8.0-SNAPSHOT</version>
```
andreaturli: remove this version

planner/optimizer/optimizer-core/pom.xml
```diff
@@ -0,0 +1,31 @@
...
+  <name>optimizer-core</name>
+  <url>http://maven.apache.org</url>
+  <properties>
+    <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
+  </properties>
+  <dependencies>
+
+
```
andreaturli: weird indentation

planner/optimizer/pom.xml
```diff
@@ -18,27 +18,27 @@
...
-		</dependency>
 		<dependency>
 			<groupId>org.slf4j</groupId>
 			<artifactId>slf4j-api</artifactId>
 		</dependency>
 		<dependency>
-			<groupId>org.testng</groupId>
-			<artifactId>testng</artifactId>
-			<scope>test</scope>
+			<groupId>org.slf4j</groupId>
```
andreaturli: remove this dependency: see https://github.com/SeaCloudsEU/SeaCloudsPlatform/commit/872e6c5d3ff6f2eddc9e06dd4a6c6547390f5918

planner/pom.xml
```diff
@@ -44,7 +44,7 @@ limitations under the License.
 
     </dependencies>
 
-    <repositories>
+     <repositories>
```
andreaturli: is it still needed?

rosogon:
@perezp : It looks like optimizer-core and src have the same files. I guess you copied the files from src to optimizer-core and forgot to remove in src. If this was your intention, the good state is that git realizes you are moving the files, so they don't appear as new content. The PR would be much easier to review, because it will only contain the service code.

perezp:
@rosogon Thank you for the comment! 

I did my best with eclipse using only the "move" and "rename" options... but something could have failed or I made also some raw copy. Actually I didn't know it was duplicated. One of the parts is useless because it is code that is in a project with "pom" packaging.

planner/optimizer/optimizer-core/pom.xml
perezp: Done, thanks!
Change not pushed yet.

planner/optimizer/optimizer-core/pom.xml
perezp: Done. Thanks!

planner/optimizer/pom.xml
perezp: Done, Thanks!

planner/pom.xml
perezp: It was not my intention to change the pom.xml of planner. It looks like a white space was accidentally included.

perezp:
@andreaturli @rosogon 
I've done the changes you have proposed. Thank you very much. All of them were necessary. 
I've already pushed the new version including your suggestions. Let me know.

andreaturli:
@perezp I think all the yaml file like `src/test/java/eu/seaclouds/platform/planner/optimizerTest/resources/MMoutputCleanYAMLV-15-09.yml` belong to `src/test/resources`: they are not java tests so better to use the usual maven layout. 

wdyt?

perezp:
@andreaturli Fine, I change it. I tried to include everything I did under some "..../planner/optimizer/...." path to avoid mixing things. Maybe it was not the right decision. I replace it with the usual maven layout as you suggest.

andreaturli:
thanks @perezp

PR #124, opened 2015-10-16
fix installation of platform from root

perezp:
FIX: a change in optimizer-service/pom.xml to be able to find the path of the service during the integration-test process


andreaturli:
thx

PR #125, opened 2015-10-17
Feature/new-data-collectors

MicheleGuerriero:
This PR adds the NURO and reconfiguration data collector after some code refactoring in order to better exploit Tower 4Clouds capabilities. Both the data collectors need the following environmental variables to work:

MANAGER_IP: the IP of Tower 4Clouds monitoring manager
MANAGER_PORT: the port of Tower 4Clouds monitoring manager
NURODC/DEPLOYERDC_CONFIG_FILE: the path to a ,properties configuration file

The NURO data collector collects metrics from NURO sensor.php and feed Tower 4Clouds with these. It has been developed mainly to monitor the NURO case study at both PaaS and IaaS level. 

The reconfiguration data collector collects metrics from the Deployer about the deployed applications and feed Tower 4Clouds with these. Right now it is able just to retrieve the status of the applications. The reconfiguration data collector is part of the SeaClouds platform  and need to be deployed with this. In mainly realizes the connection between the Deployer and the Monitor. A further PR is required to add a Brooklyn blueprint for this data collector to be added to the SeaClouds platform blueprint.


andreaturli:
Please copy the description of the PR in the readme.md

andreaturli:
Thx @MicheleGuerriero

I've quickly looked at the PR. Some initial comments, very minor.

I'll tested asap so that we can merge it.

andreaturli:
@MicheleGuerriero I think you can address the comments on this PR before merging it.

monitor/nuro-data-collector/README.md
```diff
@@ -0,0 +1,57 @@
...
+    NUROServerLastMinuteAverageRunTime
+    NUROServerLastMinuteAverageThroughput
+    NUROServerLastMinutePlayerCount
+    NUROServerLastMinuteRequestCount
+    NUROServerLastTenSecondsAverageRunTime
+    NUROServerLastTenSecondsAverageThroughput
+    NUROServerLastTenSecondsPlayerCount
+    NUROServerLastTenSecondsRequestCount
+
+Finally here is an example of a monitoring rule to install in order to get data from the NURO Data Collector:
```
andreaturli: where those monitoring-rules need to be installed? how?

monitor/nuro-data-collector/src/main/java/eu/seaclouds/monitor/nuroDc/MainDc.java
```diff
@@ -0,0 +1,83 @@
...
+
+        // load default values
+        mainInstance.loadDefaultValues();
+
+        // try to load from environment variables
+        mainInstance.loadFromEnrivonmentVariables();
+
+        // load properties from the config file
+        try {
+            nuroDCProp.load(new FileInputStream(mainInstance.configFilePath));
```
andreaturli: what is the semantic of these different config strategies?

You are using `environment variables` and a config file. Would it be possible to use just one mechanism? I think this would simplify the bootstrap of the DC and its runtime management. Wdyt?

MicheleGuerriero: The decision of having both the mechanism was because the currently required environmental variables are the same for each Data Collector, while in the .properties file there are DC-specific additional information, but of course the two mechanism for getting inputs can be merged. In this case I would prefer using just environmental variables, but I'm not sure about this. What do you suggest?

MicheleGuerriero:
Ok, thank @andreaturli, I will work on your suggestions and update the PR.

andreaturli:
I think `env var` are ok but if 2 DCs will be deployed on the same host, we may have problems.
Maybe a config file is safer

andreaturli:
thanks @MicheleGuerriero

PR #127, opened 2015-10-19
Removed deprecated Optimizer  dependency from the dashboard.

adriannieto:
@andreaturli could you review this?

### DIFF
--- a/dashboard/pom.xml
+++ b/dashboard/pom.xml
@@ -33,12 +33,6 @@
     <dependencies>
         <dependency>
             <groupId>eu.seaclouds-project</groupId>
-            <artifactId>optimizer</artifactId>
-            <version>${project.version}</version>
-        </dependency>
-        
-        <dependency>
-            <groupId>eu.seaclouds-project</groupId>
             <artifactId>aamwriter</artifactId>
             <version>${project.version}</version>
         </dependency>
--- a/dashboard/src/main/java/eu/seaclouds/platform/dashboard/resources/PlannerResource.java
+++ b/dashboard/src/main/java/eu/seaclouds/platform/dashboard/resources/PlannerResource.java
@@ -18,74 +18,23 @@
 package eu.seaclouds.platform.dashboard.resources;
 
 
-import com.google.gson.JsonObject;
 import eu.seaclouds.platform.dashboard.config.PlannerFactory;
-import eu.seaclouds.platform.dashboard.http.HttpPostRequestBuilder;
-import eu.seaclouds.platform.planner.optimizer.Optimizer;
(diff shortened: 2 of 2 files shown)


andreaturli:
good cleaning, but what's the purpose of the PlannerResource now?

I think you can rename this PR into something like `refactor PlannerResource` where you remove `optimize` and `matchmake` operations but you add `POST generateFeasiblePlans` and ` POST submitPlan`

In fact, I think `plan` is not enough as the planner should generate a number of sensible plans that the user can review and pick. Makes sense?

adriannieto:
My original idea was to directly integrate the new planner version via rest proxy, but sadly I found some issues while I was deploying it with the Catalog. They probably are related with a bad wiring of the config parameters (like #121). As part of this integration I had to remove the non working code on the dashboard side in order to avoid more dependency conflicts around this topic but as now the integration is on hold, I decided to open a PR with just the dependency fix.

Do you suggest to close this PR until the planner will be working properly and then open a new one providing fully integration with the planner?

About the call signature, the planner (due I made a bad suggestion sadly) uses GET requests (this is straightforward to be modified) instead of POST do you think that those calls should be modified into POST requests? (for me passing a file thought urlencoded params it's insane, but sadly when I made the suggestion I didn't realzied that GET doesn't support body params)

andreaturli:
I think you don't need to close this PR. 
The Dashboard needs to use the API exposed by the underlying services. If one of the consumed services is not working yet, we will need to fix that service itself.

For the signature, well, this is the semantic of RESTful services, I think. But we may convert GET to POST or viceversa later, if needed

adriannieto:
Nice  I will mock the call, as it were functional instead of delaying the development.

rosogon:
This fixes the problem I have running the dashboard.

andreaturli:
@adriannieto I suggest to look at https://github.com/square/okhttp/tree/master/mockwebserver

andreaturli:
With the current version of the dashboard, I can see
```
$ cat console.out
Exception in thread "main" java.lang.ExceptionInInitializerError
Caused by: java.lang.ClassCastException: org.slf4j.impl.Log4jLoggerAdapter cannot be cast to ch.qos.logback.classic.Logger
	at io.dropwizard.logging.LoggingFactory.bootstrap(LoggingFactory.java:43)
	at io.dropwizard.logging.LoggingFactory.bootstrap(LoggingFactory.java:37)
	at io.dropwizard.Application.<clinit>(Application.java:20)
```
so I think this fix is absolutely needed!

adriannieto:
@andreaturli so maybe then we can merge this and then add the real planner in a former step? In this moment the dashboard has a skip button in step three so the planner step could be skipped.

andreaturli:
ok, meanwhile add the support for `plan`, please.

PR #128, opened 2015-10-19
Simple TOSCA topology for deploying a TomcatServer on aws

kiuby88:
A first TOSCA topology supported on brooklyn-tosca project.


andreaturli:
thanks @kiuby88

PR #129, opened 2015-10-19
Cosmetic improvements on the dashboard step 2

adriannieto:
This removes the ugly button on the dashboard step two. @rosogon  could you take a look?


rosogon:
+1.

As an improvement, we could have a "palette" with the icons instead of the text (and a tooltip indicating the type name)

PR #130, opened 2015-10-20
increased debug verbosity and adding node templates in optimizer output

perezp:
The information of cloud resources was not included in the output information of the optimizer, only the names of the cloud resources. 
This PR adds such information to the ADPs created by the optimizer.


planner/optimizer/optimizer-core/src/main/java/eu/seaclouds/platform/planner/optimizer/OptimizerInitialDeployment.java
```diff
@@ -36,11 +36,12 @@
 
 public class OptimizerInitialDeployment {
 
+   private static final String NL = System.getProperty("line.separator");
+
```
kiuby88: I found that Java 7 has `System.lineSeparator()` too.

perezp: Changed to System.lineSeparator() . 
Thank you!

andreaturli:
lgtm

PR #133, opened 2015-10-26
Adding Python support

kiuby88:


adriannieto:
It looks nice as a first approach. Does it contain service support?

kiuby88:
Yes, it is. All CfWebApp entities inherit from [CloudFoundryWebApp](https://github.com/SeaCloudsEU/SeaCloudsPlatform/blob/5a1d0b7624ce9e7539cdd3814054b6bb85fcdaca/deployer/src/main/java/org/apache/brooklyn/entity/cloudfoundry/webapp/CloudFoundryWebApp.java) which defines a generic service support.

adriannieto:
Great, LGTM

PR #134, opened 2015-10-26
Fix. Avoiding not-enable services will be able to be accessed.

kiuby88:


rosogon:
Hi @kiuby88 . This PR is still valid, right? We have been testing it to deploy our case study and it works. It's +1 for me, even if it needs to be improved.

kiuby88:
Yes @rosogon. It PR is necessary to deploy an application with bound services on CF, such as Atos use case.

rosogon:
Ok, merging

kiuby88:
Thx

PR #138, opened 2015-11-02
Optimizer in BOM catalog for BYON deployment

perezp:
This PR changes the BOM catalog file in order to install the optimizer when the platform is deployed on BYON. 

There are just a few changes, but they may involve several people.  
@szenzaro I think that modified line 16 is interesting for the Planner. Is it OK? 
@andreaturli I took the option of starting a new Tomcat server for the optimizer. At first, I tried to deploy the optimizer war in the Tomcat server of the SLA, which was already running. However, later I discarded that option because we would create two different mapping of services onto servers depending on the kind of deployment (BYON or cloud), and I preferred to have only one. This deployment on a dedicated Tomcat is in lines 274-279. Is it OK? 
@rosogon Can you look at lines 274-279 too? I wrote them being heavily inspired in your specification of the sla-core .

Tests that I've done for the BYON deployment: every module in Brooklyn is green; 2 Tomcat servers are deployed in the VM named "SeaClouds Dashboard + Planner + SLA + Optimizer"; the first server listens in port 8080 and it is for the SLA, while the second server listens in 18080 and runs the optimizer service. 

For the deployment on the cloud, I would like to get first some advice from you regarding the best option  between deploying a dedicated VM for this optimizer service or deploying the service in the sam VM of other already installed service.


seaclouds-catalog.bom
```diff
@@ -13,10 +13,11 @@ brooklyn.catalog:
         planner.url: https://oss.sonatype.org/service/local/artifact/maven/redirect?r=snapshots&g=eu.seaclouds-project&a=planner-service&v=LATEST&e=jar
         planner.config.port: 1234
         planner.config.discoverer.url: undefined
-        planner.config.optimizer.url: undefined
+        planner.config.optimizer.url: $brooklyn:formatString("http://%s:%s", component("optimizer-service").attributeWhenReady("host.address"), component("optimizer-service").attributeWhenReady("http.port"))
```
adriannieto: This is the default value of the entity, I would rather set a default value like localhost:1234 and then, pass the brooklyn.format.... when you add the entity to the byon. If you want to see an example, you could take a look to the dashboard section in the bom.

perezp: @adriannieto Thank you. I see what you mean regarding the default value. I change it.

seaclouds-catalog.bom
```diff
@@ -254,6 +258,8 @@ brooklyn.catalog:
               planner.port: $brooklyn:component("planner").config("planner.config.port")
             install.latch: $brooklyn:component("planner").attributeWhenReady("service.isUp")
           - type: seaclouds-planner
+            brooklyn.config: 
+               planner.config.optimizer.url: $brooklyn:formatString("http://%s:%s", component("optimizer-service").attributeWhenReady("host.address"), component("optimizer-service").attributeWhenReady("http.port"))
```
perezp: @adriannieto Is this the correct place for the  planner.config.optimizer.url assignment?

adriannieto: Yes, you have to override the default values when you use the type. Please don't forget to update also the cloud deployment.

perezp: Thank you Adrián!!
 
I'm waiting for some comments/advices regarding the cloud deployment. My doubt was: should I place the optimizer in a dedicated VM or colocated with other modules?

adriannieto: IMO in order to reduce the amount of VM used it should be part of the planner VM

perezp: The option you propose is fine with me. 

If there are not comments against it at the end of the day, your opinion will became law and I will implement it.

szenzaro:
@perezp I do not know the catalog scripting language syntax and semantic. It seems good but I can't be sure.

PR #139, opened 2015-11-03
Feature(Planner): integration of DAM generation and Optimizer service

szenzaro:
@perezp could you please review this?

This PR adds dam generation fixes and optimizer integration for the planner.


planner/service/plannerconf.yml
```diff
@@ -5,7 +5,7 @@ server:
   connector:
     type: http
     port: 1234
-discovererURL: undefined
```
adriannieto: If I understood well are they default values just for testing on your machine? It reminds me to https://github.com/SeaCloudsEU/SeaCloudsPlatform/blob/master/dashboard/config.yml but instead of hardcoding those values, I have them wired with env variables (which are added in the launch configuration in InteliJ)

szenzaro: Right

common/src/main/java/eu/seaclouds/common/tosca/ToscaSerializer.java
```diff
@@ -55,18 +65,19 @@ public static String toTOSCA(Topology topology, String author, String name, Stri
     }
 
     public static ParsingResult<ArchiveRoot> fromTOSCA(Path filePath) throws ParsingException {
-        ToscaParser parser = new ToscaParserSupplier().get();
+        ToscaParser parser = ToscaSerializer.getParser();
```
andreaturli: Consider Suppliers.memoized, see http://docs.guava-libraries.googlecode.com/git/javadoc/com/google/common/base/Suppliers.html

planner/core/src/main/java/eu/seaclouds/platform/planner/core/HttpHelper.java
```diff
@@ -83,22 +84,48 @@ public String getRequest(String restPath, List<NameValuePair> params){
...
+    public String postInBody(String restPath, String bodyContent){
+        HttpPost post = new HttpPost(prepareRequestURL(restPath));
+        try {
+            StringEntity entity = new StringEntity(bodyContent);
+            post.setEntity(entity);
+            HttpResponse response = httpclient.execute(post);
+            String result = EntityUtils.toString(response.getEntity());
+            return result;
+        }catch (ClientProtocolException e) {
+            return null;
```
andreaturli: Don't ignore exceptions

planner/core/src/main/java/eu/seaclouds/platform/planner/core/HttpHelper.java
```diff
@@ -83,22 +84,48 @@ public String getRequest(String restPath, List<NameValuePair> params){
...
+        } catch (UnsupportedEncodingException e) {
+            return null;
```
andreaturli: Don't ignore exceptions

planner/core/src/main/java/eu/seaclouds/platform/planner/core/HttpHelper.java
```diff
@@ -83,22 +84,48 @@ public String getRequest(String restPath, List<NameValuePair> params){
...
+        } catch (IOException e) {
+            return null;
```
andreaturli: Don't ignore exceptions

planner/core/src/main/java/eu/seaclouds/platform/planner/core/HttpHelper.java
```diff
@@ -114,11 +141,53 @@ public String postRequest(String restPath, List<NameValuePair> params){
...
+                content = EntityUtils.toString(entity); //new Scanner(entity.getContent()).useDelimiter("\\Z").next();
+            }else{
+                content =response.getStatusLine().getReasonPhrase();
+            }
+
+            EntityUtils.consume(entity);
+            return new Pair<String, String>(String.valueOf(status), content);
+
+        } catch (UnsupportedEncodingException e) {
+            e.printStackTrace();
```
andreaturli: Use logger and don't ignore exceptions

planner/core/src/main/java/eu/seaclouds/platform/planner/core/HttpHelper.java
```diff
@@ -114,11 +141,53 @@ public String postRequest(String restPath, List<NameValuePair> params){
...
+            return null;
+        } catch (ClientProtocolException e) {
+            e.printStackTrace();
+            return null;
```
andreaturli: Use logger and don't ignore exceptions

planner/core/src/main/java/eu/seaclouds/platform/planner/core/HttpHelper.java
```diff
@@ -114,11 +141,53 @@ public String postRequest(String restPath, List<NameValuePair> params){
...
+        } catch (IOException e) {
+            e.printStackTrace();
+            return null;
```
andreaturli: Use logger and don't ignore exceptions

planner/core/src/main/java/eu/seaclouds/platform/planner/core/Planner.java
```diff
@@ -105,6 +108,7 @@ private String generatMMOutput(Map<String, HashSet<String>> mmResult, Map<String
         return mapper.writerWithDefaultPrettyPrinter().writeValueAsString(singleRes);
     }
 
+
     public String plan() throws ParsingException, IOException {
```
andreaturli: In general, avoid strings where other types are more appropriate.

Also why this method doesn't have any parameters?

andreaturli: Avoid strings where other types are more appropriate

andreaturli: Why this method doesn't have any parameters?
offerings, matchamaker, optimizeClient look like good candidates. This will enormously improve the testability of this method, just to mention a benefit

planner/core/src/main/java/eu/seaclouds/platform/planner/core/Planner.java
```diff
@@ -154,7 +161,7 @@ public String rePlan() throws ParsingException, IOException {
         //Optimize
         String mmOutput = "";
         try {
-            mmOutput = generatMMOutput(matchingOfferings, offerings);
+            mmOutput = generateMMOutput(matchingOfferings, offerings);
         } catch (Exception e) {
             //What should we return in case of failure?
             e.printStackTrace();
```
andreaturli: don't ignore exceptions

planner/core/src/main/java/eu/seaclouds/platform/planner/core/Planner.java
```diff
@@ -187,8 +196,7 @@ public String generateDam() {
         if(partialDam == null)
             return "error"; //TODO: need real error handling policy
 
-        List<NameValuePair> slaGenInfoParams = getReqParams(new BasicNameValuePair("dam", partialDam));
-        String slaInfo = new HttpHelper(slaGenURL).postRequest(DAM_GEN_OP, slaGenInfoParams);
+        String slaInfo = new HttpHelper(slaGenURL).postInBody(SLA_GEN_OP, partialDam);
 
         if(slaInfo == null)
```
andreaturli: consider checkNotNull from Guava library

planner/core/src/main/java/eu/seaclouds/platform/planner/core/Planner.java
```diff
@@ -167,17 +174,19 @@ public String rePlan() throws ParsingException, IOException {
...
 
     public String generateDam() {
 
-        List<NameValuePair> damInfoParams = getReqParams(new BasicNameValuePair("adp", adp));
-
-        String monitorInfo = new HttpHelper(monitorGenURL).postRequest(DAM_GEN_OP, damInfoParams);
+        String monitorInfo = new HttpHelper(monitorGenURL).postInBody(MONITOR_GEN_OP, adp);
 
         if(monitorInfo == null)
             return "error"; //TODO: check it is a good response, need error handling policy
```
andreaturli: consider checkNotNull from Guava library

planner/core/src/main/java/eu/seaclouds/platform/planner/core/Planner.java
```diff
@@ -187,8 +196,7 @@ public String generateDam() {
         if(partialDam == null)
             return "error"; //TODO: need real error handling policy
```
andreaturli: consider checkNotNull from Guava library

planner/core/src/main/java/eu/seaclouds/platform/planner/core/Planner.java
```diff
@@ -167,17 +174,19 @@ public String rePlan() throws ParsingException, IOException {
...
+        }
+        return null; //TODO: return failure
     }
 
     public String generateDam() {
 
-        List<NameValuePair> damInfoParams = getReqParams(new BasicNameValuePair("adp", adp));
-
-        String monitorInfo = new HttpHelper(monitorGenURL).postRequest(DAM_GEN_OP, damInfoParams);
+        String monitorInfo = new HttpHelper(monitorGenURL).postInBody(MONITOR_GEN_OP, adp);
```
andreaturli: Could `monitorInfo` be a parameter?

andreaturli: Why monitorInfo is a String?

planner/core/src/main/java/eu/seaclouds/platform/planner/core/HttpHelper.java
szenzaro: That is the next step. Adding uniform error handling. I was looking for some guidelines of SeaClouds error handling policies. Where can I find one?

planner/core/src/main/java/eu/seaclouds/platform/planner/core/Planner.java
```diff
@@ -259,11 +262,11 @@ public String generateSlaInfo(String partialDam, String slaServiceResponse){
         return reqParams;
     }
 
-    private Map<String, Pair<NodeTemplate, String>> getOfferingsFromDiscoverer() throws ParsingException, IOException {
+    private Map<String, Pair<NodeTemplate, String>> getOfferingsFromDiscoverer() throws  IOException {
         //Get discoverer ids
         // GET discovererurl/fetch -> ids [ "id1", ..., "idn" ]
         // POST discovererurl/fetch?oid=idx -> { code: "...", errormessage: "...", idx: "yaml"  }
-        String idsString = discovererClient.getRequest("/fetch", Collections.EMPTY_LIST);
+        String idsString = discovererClient.getRequest("discoverer/fetch", Collections.EMPTY_LIST);
```
andreaturli: Consider injecting the discoverClient into the method

planner/core/src/main/java/eu/seaclouds/platform/planner/core/HttpHelper.java
szenzaro: See above

planner/core/src/main/java/eu/seaclouds/platform/planner/core/Planner.java
```diff
@@ -296,32 +309,31 @@ public SLAInfo() {}
 
     }
 
-    class OfferJson {
+    public static class Offering {
```
andreaturli: extract those class into its own file. All the planner components may need to use this class instead of using String that needs to be parsed as `Offering` later

planner/core/src/main/java/eu/seaclouds/platform/planner/core/HttpHelper.java
szenzaro: see above

planner/core/src/main/java/eu/seaclouds/platform/planner/core/HttpHelper.java
szenzaro: Could you please elaborate on the logger usage?

For exceptions see above.

planner/core/src/test/java/DamGenTests.java
```diff
@@ -16,18 +16,132 @@
...
+import java.util.Map;
 import java.util.Scanner;
 
+import static org.testng.AssertJUnit.assertEquals;
 import static org.testng.AssertJUnit.assertNotNull;
 import static org.testng.AssertJUnit.assertTrue;
 
 public class DamGenTests {
 
     @Test
```
andreaturli: move up the @Test annotation at class level

planner/core/src/main/java/eu/seaclouds/platform/planner/core/HttpHelper.java
szenzaro: See above

planner/core/src/main/java/eu/seaclouds/platform/planner/core/HttpHelper.java
szenzaro: See above

planner/core/src/main/java/eu/seaclouds/platform/planner/core/Planner.java
szenzaro: That method generate a json String so String seems good to me.
The parameters are not needed as they are in the state of the planner.

How that improves testability?

planner/core/src/test/java/DamGenTests.java
```diff
@@ -16,18 +16,132 @@
...
+    public void damTranslation() throws Exception{
+        String adp = new Scanner(new File(Resources.getResource("example_adp.yml").toURI())).useDelimiter("\\Z").next();
+        Yaml yml =new Yaml();
+
+        ArrayList<Object> groupsToAdd = new ArrayList<>();
```
andreaturli: Don't declare variables as ArrayList - instead declare as a more generic type, such as java.util.List

planner/core/src/main/java/eu/seaclouds/platform/planner/core/Planner.java
szenzaro: see above

planner/core/src/test/java/DamGenTests.java
```diff
@@ -16,18 +16,132 @@
...
+        HashMap<String, ArrayList<String>> groups = new HashMap<>();
```
andreaturli: Don't declare variables as HashMap - instead declare as a more generic type, such as java.util.Map

planner/core/src/main/java/eu/seaclouds/platform/planner/core/Planner.java
szenzaro: Ok

planner/core/src/test/java/DamGenTests.java
```diff
@@ -16,18 +16,132 @@
...
 import java.util.Scanner;
 
+import static org.testng.AssertJUnit.assertEquals;
 import static org.testng.AssertJUnit.assertNotNull;
 import static org.testng.AssertJUnit.assertTrue;
 
 public class DamGenTests {
 
     @Test
+    public void damTranslation() throws Exception{
```
andreaturli: Add a javadoc to explain this test

planner/core/src/main/java/eu/seaclouds/platform/planner/core/Planner.java
szenzaro: This will change if SeaClouds adopts a uniform error handling policy. But ok guava is an option.

planner/core/src/main/java/eu/seaclouds/platform/planner/core/Planner.java
szenzaro: see above

planner/core/src/test/java/PlannerClient.java
```diff
@@ -34,7 +34,7 @@ public static void main(String[] args) throws IOException {
         String content = new HttpHelper("http://localhost:8080").getRequest("/plan", params);
         System.out.println(content);
 
-        content = new HttpHelper("http://localhost:8080").postRequest("/plan", params);
+        content = new HttpHelper("http://localhost:8080").postRequest("/plan", params).second;
         System.out.println("POST!!!! " + content);
```
andreaturli: Use a logging framework, and always log key events

planner/core/src/main/java/eu/seaclouds/platform/planner/core/Planner.java
szenzaro: monitorInfo could be a parameter, but since it is used just there there is no need to promote it. Any ideas?
The generator returns a JSon String, since. Non need to add a specific POJO for that.

planner/core/src/main/java/eu/seaclouds/platform/planner/core/Planner.java
szenzaro: Yes, you are right.

planner/core/src/test/java/PlannerTest.java
```diff
@@ -0,0 +1,49 @@
...
+import static org.testng.AssertJUnit.assertEquals;
+
+@Test(groups = {"Live"})
+public class PlannerTest {
+
+    public void optimizerIntegration() throws Exception{
+        String aam = new Scanner(new File(Resources.getResource("aam_opttest.yml").toURI())).useDelimiter("\\Z").next();
+        String mmOutput = new Scanner(new File(Resources.getResource("mmOutputExample.json").toURI())).useDelimiter("\\Z").next();
+
+        HttpHelper http = new HttpHelper("http://planner-seacloudspolimi.rhcloud.com/");
```
andreaturli: what if this URL is not available?

planner/core/src/test/java/PlannerTest.java
```diff
@@ -0,0 +1,49 @@
...
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Scanner;
+
+import static org.testng.AssertJUnit.assertEquals;
+
+@Test(groups = {"Live"})
+public class PlannerTest {
+
+    public void optimizerIntegration() throws Exception{
```
andreaturli: What is suppose to test this method? Please add some javadoc to explain the intention as it is not clear.
As it is, it seems a OptimizerClientTest not a PlannerTest then.

planner/core/src/main/java/eu/seaclouds/platform/planner/core/HttpHelper.java
andreaturli: Don't ignore exceptions

planner/core/src/main/java/eu/seaclouds/platform/planner/core/HttpHelper.java
andreaturli: Don't ignore exceptions

planner/core/src/main/java/eu/seaclouds/platform/planner/core/Planner.java
szenzaro: Not really. It is inside Planner since it is used by this class. Scattering inner classes in actual different file does not seem to be a god practice, although Java encourage to do that.

planner/core/src/main/java/eu/seaclouds/platform/planner/core/HttpHelper.java
andreaturli: Don't ignore exceptions

planner/core/src/test/java/DamGenTests.java
szenzaro: Ok

planner/core/src/main/java/eu/seaclouds/platform/planner/core/HttpHelper.java
andreaturli: Don't ignore exceptions

planner/core/src/main/java/eu/seaclouds/platform/planner/core/HttpHelper.java
```diff
@@ -114,11 +141,53 @@ public String postRequest(String restPath, List<NameValuePair> params){
...
+            }else{
+                content =response.getStatusLine().getReasonPhrase();
+            }
+
+            EntityUtils.consume(entity);
+            return new Pair<String, String>(String.valueOf(status), content);
+
+        } catch (UnsupportedEncodingException e) {
+            e.printStackTrace();
+            return null;
```
andreaturli: Don't ignore exceptions

planner/core/src/main/java/eu/seaclouds/platform/planner/core/HttpHelper.java
andreaturli: Don't ignore exceptions

planner/core/src/main/java/eu/seaclouds/platform/planner/core/HttpHelper.java
```diff
@@ -114,11 +141,53 @@ public String postRequest(String restPath, List<NameValuePair> params){
...
+        } catch (IOException e) {
+            e.printStackTrace();
+            return null;
+        }finally {
+            try {
+                response.close();
+            } catch (IOException e) {
+                e.printStackTrace();
+                return new Pair<>("500", "Post Exception for: " + httpPost.toString());
```
andreaturli: Is it really needed?

planner/core/src/main/java/eu/seaclouds/platform/planner/core/Planner.java
```diff
@@ -117,23 +121,26 @@ public String plan() throws ParsingException, IOException {
...
-        //return result
-        return optResult; //TODO: check the output for the optimizer
+
+        Pair<String, String> optResult = optimizerClient.postRequest("optimizer/optimize", optParams);
+
+
+        if(optResult.first.equals("200")){
+            return optResult.second; //TODO: check the output for the optimizer
+        }
+        return null; //TODO: return failure
```
andreaturli: https://github.com/google/guava/wiki/UsingAndAvoidingNullExplained

planner/core/src/test/java/DamGenTests.java
szenzaro: But I need exaclty an arrayList, not a generic list. If you think this will improve I will change it.
This method is an attempt to translate the ADP to Jose's DAM. But this will change since the DAM definition is WIP.

planner/core/src/main/java/eu/seaclouds/platform/planner/core/Planner.java
```diff
@@ -167,17 +174,19 @@ public String rePlan() throws ParsingException, IOException {
...
-        return optResult; //TODO: check the output for the optimizer
+
+        Pair<String, String> optResult = optimizerClient.postRequest("/optimize", optParams);
+        if(optResult.first.equals("200")){
+            return optResult.second; //TODO: check the output for the optimizer
+        }
+        return null; //TODO: return failure
     }
 
     public String generateDam() {
```
andreaturli: Avoid strings where other types are more appropriate

planner/core/src/main/java/eu/seaclouds/platform/planner/core/Planner.java
```diff
@@ -167,17 +174,19 @@ public String rePlan() throws ParsingException, IOException {
...
+                new BasicNameValuePair("offers", mmOutput)
         );
-        String optResult = optimizerClient.postRequest("/optimize", optParams);
-        return optResult; //TODO: check the output for the optimizer
+
+        Pair<String, String> optResult = optimizerClient.postRequest("/optimize", optParams);
+        if(optResult.first.equals("200")){
+            return optResult.second; //TODO: check the output for the optimizer
+        }
+        return null; //TODO: return failure
```
andreaturli: https://github.com/google/guava/wiki/UsingAndAvoidingNullExplained

planner/core/src/main/java/eu/seaclouds/platform/planner/core/Planner.java
szenzaro: The optimizer returns a JSon string that I must forward to the service caller. So String is more appropriate than a class to be serialized again.

planner/core/src/main/java/eu/seaclouds/platform/planner/core/Planner.java
szenzaro: See above

planner/core/src/main/java/eu/seaclouds/platform/planner/core/HttpHelper.java
andreaturli: @szenzaro what do you mean by SeaClouds error handling? HttpHelper is a util class that could define its contract better, I think. Chapter 8 - Effective Java Joshua Block may give you a good idea

szenzaro: I mean, SeaClouds is mad of a bunch of services. What is the expected behaviour in case of failure of this services. Is there an agreed code-style, error handling and notification policy or something similar?  Should each developer do as he like?

planner/core/src/main/java/eu/seaclouds/platform/planner/core/HttpHelper.java
szenzaro: yes

planner/core/src/main/java/eu/seaclouds/platform/planner/core/HttpHelper.java
szenzaro: see above

planner/core/src/main/java/eu/seaclouds/platform/planner/core/HttpHelper.java
andreaturli: HttpHelper is not a service. For what I can see HttpHelper is yet another util class that tries to simplify Apache HttpClient. 

Don't try to solve all the problems at once, try to improve the implementation of this class first.

planner/core/src/test/java/PlannerTest.java
szenzaro: Ok, I will change the class name so it will be more clear.

planner/core/src/main/java/eu/seaclouds/platform/planner/core/HttpHelper.java
andreaturli: http://stackoverflow.com/questions/2750346/logger-vs-system-out-println

planner/core/src/test/java/PlannerTest.java
szenzaro: The test will fail

planner/core/src/test/java/PlannerClient.java
szenzaro: Ok

planner/core/src/main/java/eu/seaclouds/platform/planner/core/Planner.java
andreaturli: http://misko.hevery.com/attachments/Guide-Writing%20Testable%20Code.pdf

planner/core/src/test/java/DamGenTests.java
szenzaro: This is a piece of code that is WIP. It is not a real test, I will remove it since it seems to generate confusion to you.

planner/core/src/test/java/DamGenTests.java
szenzaro: ok

planner/core/src/main/java/eu/seaclouds/platform/planner/core/HttpHelper.java
szenzaro: See above

planner/core/src/main/java/eu/seaclouds/platform/planner/core/HttpHelper.java
szenzaro: See above

planner/core/src/main/java/eu/seaclouds/platform/planner/core/HttpHelper.java
szenzaro: See above

planner/core/src/main/java/eu/seaclouds/platform/planner/core/HttpHelper.java
szenzaro: See above

planner/core/src/main/java/eu/seaclouds/platform/planner/core/Planner.java
andreaturli: I'm talking about something like

```
import it.polimi.tower4clouds.rules.MonitoringRules;

DAM generateDAM(MonitorRules monitorRules) { ... }
```
or similar. wdyt?

planner/core/src/main/java/eu/seaclouds/platform/planner/core/Planner.java
andreaturli: A public inner class will be accessible be the outer class so no reason to be defined as nested class.
In fact, `Offering` can be reused by Matchmaker, Discoverer and possibly Optimizer, just to mention the first components that come to my mind: I hope you see why it is conceptually not an inner class.

planner/core/src/test/java/DamGenTests.java
andreaturli: I'm referring to the declaration not the creation of the list

planner/core/src/test/java/DamGenTests.java
andreaturli: I don't want to be a problem. I'm simply saying it is not clear and a javadoc could help here.

planner/core/src/main/java/eu/seaclouds/platform/planner/core/Planner.java
andreaturli: If the optimizer returns a String (why a json? I think planner wants to use the optimizer as a lib, not as a service?) I think Optimizer can be improved as well.
Anyway, `Planner.generateDam()` output doesn't need to be a String because Optimizer.optimize output is a String, don't you think?

planner/core/src/main/java/eu/seaclouds/platform/planner/core/HttpHelper.java
rosogon: @szenzaro : At several points of the code, you are returning null at several points when an exception is caught. If you return (instead of rethrowing the exception), the exception must be logged. Otherwise, you won't know what is happening, even when testing the application. 

This hasn't to do with having a error-handling policy or not.

szenzaro: Ok

planner/core/src/main/java/eu/seaclouds/platform/planner/core/Planner.java
andreaturli: @szenzaro leave this improvement for a subsequent PR, if needed at all.

andreaturli:
Thanks @szenzaro 
In summary, I think it would be good if you can address my comments on exceptions handling and logger usage. No need to refactor all of the planner methods to return something different from String in this PR.
HTH

planner/core/src/main/java/eu/seaclouds/platform/planner/core/Planner.java
andreaturli: See https://github.com/SeaCloudsEU/SeaCloudsPlatform/blob/master/discoverer/src/main/java/eu/seaclouds/platform/discoverer/core/Offering.java

szenzaro: Those are two different domains, and that class in the discoverer is changed (and will be PR soon)

szenzaro:
@andreaturli ok, I will address your comments ASAP

planner/core/src/main/java/eu/seaclouds/platform/planner/core/Planner.java
andreaturli: This is exactly the problem I'm trying to avoid here. We should have a single domain object that represents the same concept.

adriannieto:
@andreaturli @szenzaro It seems that with this PR the planner is not wired in the seaclouds-catalog, maybe for the next one you should focus on providing a unified installation experience.

szenzaro:
@andreaturli could you please check why it fails on:

The command "aws s3 sync s3://seaclouds-m2/repository/ /home/travis/.m2/repository/ --quiet" failed and exited with 1 during .

andreaturli:
@szenzaro , please rebase your branch on push it again, it should be ok now.

szenzaro:
I rebased the PR. Now it should be ok.

perezp:
I feel perplexed by the fact that I do not see not tests for the Plan method of the Planner yet. 
I merge this PR because I see that now logger is used now and Exceptions do not return null directly.
Tests are a must for the next PR.

PR #142, opened 2015-11-05
Modified Dashboard Dropwizard configuration in order to use just one port

adriannieto:
@rosogon could you review it?


rosogon:
Just to know, why do you want to change this?

adriannieto:
I think it's better to reduce the port usage to avoid collisions.

rosogon:
+1. Merging

PR #140, opened 2015-11-05
Wired Dashboard with the Planer for automatic ADP generation. Small fixes.

adriannieto:
This PR integrates the planner with the dashboard. Currently there is no  way to test the functionality by using BOM due a problem with the catalog. Please the planner endpoint to http://131.114.3.217:1234 if you want to test it.

@rosogon could you review this?

PR #144, opened 2015-11-06
DAM Generator partially wired to the dashboard

adriannieto:
This allows the step 4 to autogenerate the DAM based on the selected ADP. Currently it doesn't feature the automatic generation of the SLA  Agreements and the Monitoring Rules.

Please notice that the Planner it's not wired properly in the catalog, please use an external planner IP in order to test it. (Same as #140  http://131.114.3.217:1234)

@rosogon  could you take a look?

Thank you!


rosogon:
+1

PR #145, opened 2015-11-09
FIX(dashboard): Nodes behaviour was lost when going back to step 2

rosogon:


rosogon:
Hi, @perezp. ¿Could you have a look at why this PR is failing?

adriannieto:
Thank you @rosogon nice catch. It seems that you fixed partially this annoying bug, however it looks that only the edit button is working, because when you back to the step 2 from the step 3, you can't delete nodes. Could you take a look?

perezp:
@rosogon Done. 
External service hosting the optimizer that is requested in some tests was not active. 
I'll remove this test to avoid more problems of this type in the future.

adriannieto:
@rosogon Please do the following, goto step 2, add a node, go to step 3, back to step 2 remove a node, add a node. You can't remove this one. Could you check it?

adriannieto:
Nice, now it works! Just squash it before merge. +1

rosogon:
Squashed. Merging

PR #150, opened 2015-11-11
Added the monitoring-dam-generator and the monitoring-dam-generator-s…

MicheleGuerriero:
This PR adds the monitoring-dam-generator and the monitoring-dam-generator-service. The latter basically wraps the former in order to make it available as a service. Both the components have their own README.md in which you can find details on how they work and how they can be used. Mainly they allow to generate given an application ADP, a the set of monitoring rules and a set of data collector deployment scripts required for that application monitoring. The generation and management of IDs associated to the generated artifacts is the only feature specific to the service. If needed in the future, this could be moved to the core. The data collector TOSCA deployment scripts are generated according to the current DAM definition ( https://docs.google.com/document/d/1Z9RdaWn54IomgI-jCiw-tdutsCdDoHfk67PNUASOFmY/edit ).

Issues:

- the monitoring-dam-generator-service currently does not store the generated information in any persistent layer, but it just keep the state in memory. A next development will be to add a persistence solution.

- the generated bash script required for the deployment of data collector are currently not uploaded anywhere, since we didn't define a remote repository for this kind of artifacts.


monitor/monitoring-dam-generator-service/README.md
```diff
@@ -0,0 +1,30 @@
...
+POST /damgen : this is the main endpoint which generates the application specific monitoring information. The request body must be the yaml representing the application ADP. It returns a JSON object with a single field containing an ID which can be used later on in order to retrieve from the Monitoring DAM Generator Service itself the generated rules and data collectors deployment scripts.
+
+GET /{applicationId}/generatedDam : this interface returns a JSON object in which each field represent a TOSCA based deployment script for one of the data collectors which have to be deployed with the application.
+
+GET /{applicationId}/monitoringRule : this interface return a JSON object with a single field contaning all the monitoring rules that have to be installed in order to monitor the application.
+
+In order to run the service one need to use the following lines of code:
+
+    export MONITOR_IP=<MONITORING-MANAGER-IP>    
+    export MONITOR_PORT=<MONITORING-MANAGER-PORT>
```
andreaturli: Try to use the config.yaml to inject those values

monitor/monitoring-dam-generator-service/README.md
```diff
@@ -0,0 +1,30 @@
...
+    export MONITOR_IP=<MONITORING-MANAGER-IP>    
+    export MONITOR_PORT=<MONITORING-MANAGER-PORT>    
+    cd monitoring-dam-generator    
+    mvn clean install    
+    cd target    
+    java -jar monitoring-dam-generator-service-0.8.0-SNAPSHOT.jar server ../config.yaml 
+
+# Issues
+
+Currently the Monitoring DAM Generator Service is not storing the generated application IDs along with the rules and data collectors script in a persistent system. The service is basically stateful and keep all the generate dinformation in memory until it is running. Future developments will be devoted to add a persistence layer.
```
andreaturli: is it an issue? :)

monitor/monitoring-dam-generator-service/pom.xml
```diff
@@ -0,0 +1,83 @@
...
+    </dependency>
+    
+    <dependency>
+        <groupId>org.testng</groupId>
+        <artifactId>testng</artifactId>
+        <scope>test</scope>
+    </dependency>   
+  </dependencies>
+  
+  <build>
```
andreaturli: why this plugin?

monitor/monitoring-dam-generator-service/src/main/java/eu/seaclouds/monitor/monitoringDamGeneratorService/ApplicationRules.java
```diff
@@ -0,0 +1,20 @@
+package eu.seaclouds.monitor.monitoringDamGeneratorService;
```
andreaturli: weird package name

monitor/monitoring-dam-generator-service/src/main/java/eu/seaclouds/monitor/monitoringDamGeneratorService/Service.java
```diff
@@ -0,0 +1,118 @@
...
+        monitoringManagerPort = monitorPort;
+      
+        }
+    
+    
+    @POST
+    @Path("/damgen")
+    @Produces(MediaType.APPLICATION_JSON)
+    @Consumes(MediaType.TEXT_PLAIN)
+    public ApplicationId generateMonitoringInfo(String adp) {
```
andreaturli: I'd expect "public ApplicationId generateMonitoringInfo(ADP adp)" vs "public ApplicationId generateMonitoringInfo(String adp)"  - wdyt?

monitor/monitoring-dam-generator-service/src/main/java/eu/seaclouds/monitor/monitoringDamGeneratorService/Service.java
```diff
@@ -0,0 +1,118 @@
...
+            String monitorPort)
+        {
+        monitoringManagerIp = monitorUrl;
+        monitoringManagerPort = monitorPort;
+      
+        }
+    
+    
+    @POST
+    @Path("/damgen")
```
andreaturli: The pathname should reflect the action

monitor/monitoring-dam-generator-service/src/main/java/eu/seaclouds/monitor/monitoringDamGeneratorService/Service.java
```diff
@@ -0,0 +1,118 @@
...
+            e.printStackTrace();
+        }
+
+        return new ApplicationRules(sw.toString());
+    }
+    
+    @GET
+    @Path("/{id}/generatedDam")
+    @Produces(MediaType.APPLICATION_JSON)
+    public ApplicationToscaDcScripts getGeneratedPiecesOfDamByApplication(@PathParam("id") String applicationId){
```
andreaturli: add javadoc as the operation is not clear

monitor/monitoring-dam-generator-core/src/test/java/eu/seaclouds/monitor/monitoringDamGenerator/MainTest.java
```diff
@@ -0,0 +1,109 @@
...
+import java.nio.file.Files;
+import java.nio.file.Paths;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import org.testng.annotations.Test;
+import eu.seaclouds.monitor.monitoringDamGenerator.MonitoringDamGenerator;
+import eu.seaclouds.monitor.monitoringDamGenerator.adpParsing.Module;
+
+public class MainTest {
```
andreaturli: Give it a more meaningful name :)

monitor/monitoring-dam-generator-core/src/main/java/eu/seaclouds/monitor/monitoringDamGenerator/MonitoringDamGenerator.java
```diff
@@ -0,0 +1,249 @@
...
+            .getLogger(MonitoringDamGenerator.class);
+
+    private static String monitoringManagerIp;
+    private static String monitoringManagerPort;
+    
+    private static final String NURO_DC_ID="nuroDc";
+    private static final String JAVA_APP_DC_ID="javaAppDc";
+    private static final String MODACLOUDS_DC_ID="modacloudsDc";
+
+    public MonitoringDamGenerator(String monitorUrl, String monitorPort) {
```
andreaturli: consider java.net.URL

monitor/monitoring-dam-generator-core/src/main/java/eu/seaclouds/monitor/monitoringDamGenerator/MonitoringDamGenerator.java
```diff
@@ -0,0 +1,249 @@
...
+        
+        monitoringManagerIp = monitorUrl;
+        monitoringManagerPort = monitorPort;
+
+    }
+
+    public List<Module> generateMonitoringInfo(String adp) {
```
andreaturli: I suggest to accept ADP class rather than a String

andreaturli: Please add some javadoc as this method is doing much more than what the name declares :)

monitor/monitoring-dam-generator-core/src/main/java/eu/seaclouds/monitor/monitoringDamGenerator/MonitoringDamGenerator.java
```diff
@@ -0,0 +1,249 @@
...
+                            nuroDcDeploymentScript);
+
+                }
+
+                module.setDataCollectorBashDeploymentScripts(tempDataCollectorDeploymentScripts);
+                toReturn.add(module);
+
+            }
+
+            this.generateDamMonitoringInfo(toReturn);
```
andreaturli: not sure I get this.

monitor/monitoring-dam-generator-core/src/main/java/eu/seaclouds/monitor/monitoringDamGenerator/adpParsing/Host.java
```diff
@@ -0,0 +1,47 @@
+package eu.seaclouds.monitor.monitoringDamGenerator.adpParsing;
```
andreaturli: weird package name, avoid camelCase

monitor/monitoring-dam-generator-core/src/main/java/eu/seaclouds/monitor/monitoringDamGenerator/adpParsing/Host.java
```diff
@@ -0,0 +1,47 @@
...
+
+import it.polimi.tower4clouds.rules.MonitoringRules;
+
+import java.util.Map;
+
+public class Host {
```
andreaturli: prefer immutable Value object.

monitor/monitoring-dam-generator-core/src/main/java/eu/seaclouds/monitor/monitoringDamGenerator/adpParsing/YAMLMonitorParser.java
```diff
@@ -0,0 +1,245 @@
...
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.yaml.snakeyaml.Yaml;
+
+@SuppressWarnings("unchecked")
+public class YAMLMonitorParser {
```
andreaturli: if the core accepts ADP you don't need the parser, do you?

monitor/monitoring-dam-generator-core/src/main/java/eu/seaclouds/monitor/monitoringDamGenerator/dcScriptGenerators/DataCollectorDeploymentScriptGenerator.java
```diff
@@ -0,0 +1,23 @@
+package eu.seaclouds.monitor.monitoringDamGenerator.dcScriptGenerators;
+
+import eu.seaclouds.monitor.monitoringDamGenerator.adpParsing.Module;
+
+public interface DataCollectorDeploymentScriptGenerator {
```
andreaturli: Not sure I get the idea of the DataCollectorDeployemntScriptGenerator. My understanding is that a DC is yet another TOSCA NodeType. Similarly to any other NodeType, the DC should be defined as i.e. "seaclouds.node.datacollector.X" and so the "MonitoringDamGenerator" should simply attach the string as the deployer will do the deployment.
Am I missing something?

monitor/monitoring-dam-generator-service/README.md
MicheleGuerriero: mmm I don't know, maybe more a future improvment, or you are suggesting to completly avoid to have a persistent layer?

monitor/monitoring-dam-generator-service/pom.xml
MicheleGuerriero: removed

monitor/monitoring-dam-generator-service/src/main/java/eu/seaclouds/monitor/monitoringDamGeneratorService/Service.java
MicheleGuerriero: you are right, @szenzaro do you agree (since I think you will be the one who will call this API) if I change the pathname to /generateDamMonitoringInfo? @andreaturli is the last a valid proposal?

monitor/monitoring-dam-generator-core/src/test/java/eu/seaclouds/monitor/monitoringDamGenerator/MainTest.java
MicheleGuerriero: done

monitor/monitoring-dam-generator-core/src/main/java/eu/seaclouds/monitor/monitoringDamGenerator/MonitoringDamGenerator.java
MicheleGuerriero: The difference is that generateMonitoringInfo just geneartes bash scripts that the are pointed in the interface field of the TOSCA yaml specific of a DC. The generateDamMonitoringInfo method is generating the TOSCA yaml required for that DC. Maybe I can give the method a more meaningful name or move what generateDamMonitoringInfo is doing in the toString of the Module itself.

monitor/monitoring-dam-generator-service/src/main/java/eu/seaclouds/monitor/monitoringDamGeneratorService/Service.java
szenzaro: Yes, I agree. You can change the operation name.

andreaturli:
None of my comments are blockers IMHO. Happy to merge it as-is.

monitor/monitoring-dam-generator-core/src/main/java/eu/seaclouds/monitor/monitoringDamGenerator/adpParsing/Host.java
MicheleGuerriero: right, sorry

monitor/monitoring-dam-generator-service/README.md
MicheleGuerriero: The config.yaml is actually retrieving this values from the previously exported env vars. It looks like the following:

server:
  type: simple
  applicationContextPath: /
  connector:
    type: http
    port: 8177
monitorIp: ${MONITOR_IP}
monitorPort: ${MONITOR_PORT}

Is this the correct approach?

monitor/monitoring-dam-generator-service/src/main/java/eu/seaclouds/monitor/monitoringDamGeneratorService/ApplicationRules.java
MicheleGuerriero: fixed

monitor/monitoring-dam-generator-core/src/main/java/eu/seaclouds/monitor/monitoringDamGenerator/adpParsing/Host.java
MicheleGuerriero: can I consider this https://docs.oracle.com/javase/tutorial/essential/concurrency/imstrat.html as a valid guide line for building immutable objects? SHould this apply also to the Module class?

MicheleGuerriero:
@andreaturli  Thanks a lot for the comments, I updated the PR trying to fix almost all of them. Looking at the comments, the only big issue that still I have to solve is the one about the usage of immutable objects. Since this fix would quite vastly affect the code in different points, but not its functionalities, do you think it is reasonable if I apply it with a further PR? If you agree on this and if you don't see any other issue that need to be solved within this PR, can we merge it, in order also to make the component available to the DAM Generator?

andreaturli:
@MicheleGuerriero immutable objects were just a [minor] comments. 

Can you rebase your branch and push it again so that travis will run again.

if travis is still green after the re-build, I'm happy to merge it.
Thanks!

monitor/monitoring-dam-generator-service/README.md
andreaturli: @MicheleGuerriero I'm not totally sure I've got your question but I think this one can help http://www.dropwizard.io/0.9.1/docs/manual/core.html#environment-variables

adriannieto: @andreaturli If I understood well, @MicheleGuerriero is saying is the config.yml is retrieving the parameters from env variables, so they are needed to run the dam-generator. 

This is not very precise as an user could replace the config.yml and write the endpoints manually, maybe he could rewrite the readme.md to tell the user that the dam-generator requires to setup the config.yml no matter if the variables come from env (by default) or the user input.

MicheleGuerriero:
I fixed the issue with the Dropwizard service configuration and rebased. Should I also squash the commit?

andreaturli:
yes please

MicheleGuerriero:
I squashed.

andreaturli:
Thanks @MicheleGuerriero

PR #151, opened 2015-11-13
FIX: Optimizer reduce verbosity and treatment of logs

perezp:
Current version of Optimizer in master is with high verbosity level. 
This PR reduces it and removes variables in the code that informed about expected verbosity. 

@rosogon can you look at it? 
@andreaturli I hope that travis is happy with the amount of logs now. Let me know if it is OK if you can.


planner/optimizer/optimizer-core/src/test/resources/log4j.xml
```diff
@@ -4,15 +4,51 @@
 	xmlns:log4j='http://jakarta.apache.org/log4j/'>
 
 	<appender name="console" class="org.apache.log4j.ConsoleAppender">
```
andreaturli: do you think it is possible to use something like https://wiki.apache.org/logging-log4j/FileAppender ?
Otherwise happy to merge it as-is.

perezp: Done as you suggested. Thanks.

andreaturli:
Thanks @perezp

PR #152, opened 2015-11-13
FEATURE(aamwriter): Language and not handled properties are stored in node template props

rosogon:
All the properties defined in a designer node that are explicitly handled in the translation process are stored in the properties of the node template.

There are two exceptions:

* The language of a module is implicitly stored in the node_type, just for the matchmaking process. With this PR, the language is also a property of the node template, so each process that needs to know  the languages of the components, is able to retrieve it (e.g. monitoring rules generator)
* The QoS section of each component are not stored in the node_template properties.


rosogon:
Hi @andreaturli , the build failed with the following error: 
`The command "aws s3 sync s3://seaclouds-m2/repository/ /home/travis/.m2/repository/ --quiet" failed and exited with 1 during`

I guess it is related with the QA Module. Could you have a look? Thanks.

andreaturli:
@rosogon, please rebase your branch on push it again, it should be ok now

MicheleGuerriero:
+1

PR #158, opened 2015-11-23
Fix missing grunt-cli as devDependency + Reduce verbosity on grunt build

adriannieto:


andreaturli:
@adriannieto we are saving ~250 lines , I'd hope something more, but better than nothing.

Could you also update the doc to specify that we need node.js installed to build seaclouds from source?

Thanks

PR #171, opened 2015-12-02
Simple repository

rosogon:
This is a simple implementation of an in-memory repository for storing TOSCA files, artifacts, etc.

Adding persistence to file is straightforward.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -29,6 +29,7 @@
         <module>monitor</module>
 	<module>dashboard</module>
         <module>sla</module>
+        <module>repository</module>
         <module>qa</module>
     </modules>
     <packaging>pom</packaging>
@@ -559,6 +560,11 @@
             <dependency>
                 <groupId>io.dropwizard</groupId>
                 <artifactId>dropwizard-assets</artifactId>
+                <version>${dropwizard.version}</version>
+            </dependency>
+            <dependency>
+                <groupId>io.dropwizard</groupId>
+                <artifactId>dropwizard-forms</artifactId>
                 <version>${dropwizard.version}</version>
             </dependency>
 
--- a/repository/README.md
+++ b/repository/README.md
@@ -1,0 +1,62 @@
+# repository
+Simple implementation of a SeaClouds Repository.
+
+## Usage ##
+
+The normal usage of a REST service. 
+
+/data contains endpoint to upload/download data.
+
+Supported methods:
+
+###POST /data
+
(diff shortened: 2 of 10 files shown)


andreaturli:
May I ask why you haven't considered a [NoSQL](https://en.wikipedia.org/wiki/Document-oriented_database) solution already available ?

rosogon:
Because when @fradandria asked for a repository two integration meetings ago, nobody replied anything. What already available solution do you suggest?

fradandria:
@andreaturli although I would agree on the technology you are proposing… Elastic search + (maybe) MongoDB. Here we are looking for something very simple to implement and integrate. Unfortunately we are aleady late...

ahgittin:
@rosogon @fradandria  Writing and supporting our own key-value store instead of using any one of hundreds available is **not** going to save us any time.  Sure it gets it off the implementer's plate put it makes an untenable burden for the integrators and maintainers.

-1

for the following reasons:
* Crazy idea to write our own KV store.
* Also this does not have any tests.
* And at runtime this does not provide persistence.

rosogon:
@andreaturli : I had not seen you had edited your comment. 

elasticsearch is close to our needs, but it looks it only handles json documents, and we also need to store binary files. Are you sure elasticsearch is suitable for us? Will you be in charge of preparing the bom to launch this (or other) KV store so the other developers can use it?

andreaturli:
[Apache Brooklyn](https://brooklyn.apache.org/) supports many nosql already: cassandra, couchbase, couchdb, crate, elasticsearch, mongo, redis, riak just to name some. For a complete list please have a look at [Apache Brooklyn catalog](https://brooklyn.apache.org/learnmore/catalog/index.html) and filter for `nosql`. 
Pick one of them with a good java and javascript clients, maybe so we'll have for free a good KV store already supported by [Apache Brooklyn](https://brooklyn.apache.org/). wdyt? Sounds good?

fradandria:
@andreaturli @ahgittin @rosogon Actually this is the issue with elasticsearch! AFAIK elasticsearch requires transform files in base64 encode before to store them. elasticsearch doesn't even allow to search binary files. I see we should use elasticsearch for storing AAL and DAM and Mongo for binary… or just MongoDB in any case we need to integrate them providing some logic on top…

rosogon:
Thanks, @andreaturli . 

Again:
* Which of those already supported by Brooklyn do you suggest?
* Could you integrate it into the platform?

adriannieto:
I think that in this moment, to avoid extra noise, this PR should be closed.

PR #173, opened 2015-12-02
Added user-readable ADP description in the third add-application-wizard

adriannieto:
@rosogon  could you test it?

### DIFF
--- a/dashboard/.gitignore
+++ b/dashboard/.gitignore
@@ -4,3 +4,4 @@ target
 **/webapp/static/lib/*
 !**/webapp/static/lib/angular-topology-editor/
 !**/webapp/static/lib/format-validator/
+!**/webapp/static/lib/adp-prettifier/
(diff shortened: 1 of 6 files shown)


rosogon:
Hi @adriannieto . It's better than the current screen, but I would change, in order of importance:

* The title of each offering is: **seaclouds.Nodes.Platform - (No cost estimation)**. I guess you wanted to show here the name of the provider.

* The field "Services" show sc_req.\<modulename\>. What do you want to show here? I guess is the module name. The right name is in the node template; you are retrieving the name from the node type.

* I would place the Services field before the Properties.

* Collapsible properties

The idea is fine, and I think we can merge and improve this in a later PR.

What do you think?

adriannieto:
@rosogon let me answer you point by point:
- I cant show the provider in a pretty way, I already suggested @PaoloCifariello to change the type to a more specific one, like seaclouds.Nodes.Platform.CloudFoundry. It requires the presence of this type on the ADP.
- You're right, I can address this in next PR
- IMO this doesn't make sense, the properties belong to the location not the services.
- This would be nice, I can address later on.

As you already mentioned, we can merge it, and then open a issue with this ideas and we will work on it later.

PR #183, opened 2015-12-07
FEATURE(designer): Implements a designer view to enter the cloud credentials

rosogon:
This commit adds a view to enter the credentials. It also contains code to generate a topology from a DAM, and to inject the credentials into a DAM.

Use the following snippet to initialize the canvas with the DAM:
```
topology = Credentials.to_topology(dam);
Credentials.init(canvas);
Credentials.fromjson(topology);
canvas.restart();
```

Use the following snippet to inject the credentials into the DAM:
```
dam = Credentials.store_credentials_in_dam(dam)
```

You can see an example on how to use this code here:
https://github.com/rosogon/topology-editor/blob/master/index-credentials.html#L44

@adriannieto : Please, have a look.


adriannieto:
@rosogon Are you considering to add it to the step 4 of the application wizard?

Issue #164, opened 2015-11-26
Add doc for each component


andreaturli:
Add README file to each maven module.
This will contain the following:
- Description
- Usage
- How to run it

PR #185, opened 2015-12-09
Feature/adding first brooklyn tosca support

kiuby88:
Adding a first (temporal) brooklyn support.
Brooklyn Deployer was updated to 0.9.0-SNAPSHOT version as a temporal solution too.

### DIFF
--- a/byon/files/seaclouds-catalog.bom
+++ b/byon/files/seaclouds-catalog.bom
@@ -193,6 +193,7 @@ brooklyn.catalog:
         brooklynnode.webconsole.nosecurity: true
         brooklynnode.classpath:
           - https://s3-eu-west-1.amazonaws.com/seaclouds-deployer/deployer-0.8.0-20150812.15.43.24-32-b090.jar
+          - https://s3-eu-west-1.amazonaws.com/tosca-temporal-integration/brooklyn-tosca-transformer-0.9.0-SNAPSHOT.jar
       brooklyn.enrichers:
         - enricherType: org.apache.brooklyn.enricher.stock.Transformer
           brooklyn.config:
--- a/byon/files/seaclouds-catalog.bom
+++ b/byon/files/seaclouds-catalog.bom
@@ -186,13 +186,13 @@ brooklyn.catalog:
       type: org.apache.brooklyn.entity.brooklynnode.BrooklynNode
       id: deployer
       name: Deployer
-      version: 0.8.0-incubating
+      version: 0.9.0-SNAPSHOT
 
       brooklyn.config:
         onExistingProperties: do_not_use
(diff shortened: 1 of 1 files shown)


byon/files/seaclouds-catalog.bom
```diff
@@ -186,13 +186,14 @@ brooklyn.catalog:
       type: org.apache.brooklyn.entity.brooklynnode.BrooklynNode
       id: deployer
       name: Deployer
-      version: 0.8.0-incubating
+      version: 0.9.0-SNAPSHOT
```
andreaturli: I think poiting to SNAPSHOT is not a good idea.

byon/files/seaclouds-catalog.bom
```diff
@@ -186,13 +186,14 @@ brooklyn.catalog:
...
 
       brooklyn.config:
         onExistingProperties: do_not_use
         brooklynnode.webconsole.nosecurity: true
         brooklynnode.classpath:
-          - https://s3-eu-west-1.amazonaws.com/seaclouds-deployer/deployer-0.8.0-20151007.193920-32.jar
+          - https://s3-eu-west-1.amazonaws.com/seaclouds-deployer/deployer-0.8.0-20150812.15.43.24-32-b090.jar
+          - https://s3-eu-west-1.amazonaws.com/tosca-temporal-integration/brooklyn-tosca-transformer-0.9.0-SNAPSHOT.jar
```
andreaturli: Same here, SNAPHOT is not reliable.

byon/files/seaclouds-catalog.bom
kiuby88: I agree with you, @andreaturli. However, the Deployer needs 0.9.0 brooklyn version to manage brooklyn-tosca project. So, I thought we could use SNAPSHOT version as a temporal solution.

byon/files/seaclouds-catalog.bom
kiuby88: Really, this jar contains brooklyn-tosca/brooklyn-tosca-transformer logic. So, I could rename this artifact

byon/files/seaclouds-catalog.bom
andreaturli: I think it is too risky. Let's close this PR, `seaclouds-catalog.bom` has to be deeply modified.

Focus on having a good documentation in the README.md, where you exactly specify what's needed to have the SeaClouds Deployer (Apache Brooklyn + ? ), how to install it, how to run it, etc. See https://github.com/SeaCloudsEU/SeaCloudsPlatform/issues/164

kiuby88: ok, we can close. So, I suppose that we reopen it asap, right?

PR #188, opened 2015-12-10
Feature/group management

kiuby88:
Groups are expected by Deployer as a topology_template's child.

### DIFF
--- a/monitor/monitoring-dam-generator-core/src/main/java/eu/seaclouds/monitor/monitoringdamgenerator/adpparsing/YAMLMonitorParser.java
+++ b/monitor/monitoring-dam-generator-core/src/main/java/eu/seaclouds/monitor/monitoringdamgenerator/adpparsing/YAMLMonitorParser.java
@@ -285,7 +285,7 @@ public class YAMLMonitorParser {
         try {
             for (String key : group.keySet()) {
                 if (key.equals(MEMBERS_KEY)) {
-                    return (List<String>) group.get(key);
+                    return coerceStringList((ArrayList)group.get(key));
                 }
             }
 
@@ -297,5 +297,15 @@ public class YAMLMonitorParser {
                     "The parser was not able to retrieve the 'members' field from one of the 'group' in the current ADP.");
         }
 
+    }
+
+    private List<String> coerceStringList(List list){
+        List<String> result = new ArrayList<>();
+        if(list!=null){
+            for(Object item: list){
+                result.add((String)item);
+            }
+        }
+        return result;
     }
 }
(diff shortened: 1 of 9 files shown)


kiuby88:
@szenzaro @MicheleGuerriero  please take a look to this commit. Probably, we could expect groups as a topology_template's child.

szenzaro:
+1 for me

rosogon:
@kiuby88 : This branch has conflicts. Could you resolve this?

kiuby88:
@rosogon sorry for the late response.
Rebased

MicheleGuerriero:
I think this PR will make the monitoring-dam-generator-core module to not work since it changes the format of the input and my ADP parser will fail. If it is needed to be compliant with the TOSCA specification I can fix my parser, but I have some doubts this doesn't affect only me. @kiuby88 can you please help me in clarify how to handle this point?

perezp:
@MicheleGuerriero We can talk on monday and I can make clear this change. I agree with you that this might not affect only you. 

I would keep the PR open until we solve these issues. 

@kiuby88

andreaturli:
Stop using just Strings!

andreaturli:
@kiuby88 please refactor 
```
public static String generateDam(String adp, String monitorGenURL, String monitorGenPort, String slaGenURL) {
```
into something Object Oriented like 
```
public DAM generateDAM(ADP adp) 
```
where ADP is probably just a4c's Topology and DAM is Topology + Policies, I think.

Do you think it is feasible?

adriannieto:
@kiuby88 I'm doing something similar like @andreaturli proposes with SeaCloudsApplicationData.java (in https://github.com/adriannieto/SeaCloudsPlatform/blob/feature/move-orchestration-from-js-to-java/dashboard/src/main/java/eu/seaclouds/platform/dashboard/model/SeaCloudsApplicationData.java just in case you want to take a look), I agree with @andreaturli but @szenzaro must agree on this, because maybe it will involve a lot of changes.

kiuby88:
@MicheleGuerriero , I did not know that your component used groups. I am sorry.
I have moved groups migration to `generateDam` method. 
Of course, it is a TEMPORAL solution.

kiuby88:
I have solved issues 198 and 199 on this PR as a temporal solution. These issues should be fixed from AAM generation.

kiuby88:
@szenzaro , could you take  a look?

jacopogiallo:
We (UPI) agree on the updates by @kiuby88

kiuby88:
Thanks a lot @jacopogiallo  and @andreaturli

PR #190, opened 2015-12-11
FEATURE(planner): Monitoring rule retrieval API

szenzaro:
@adriannieto since you are the one using this for the dashboard, could you please review it?

This PR adds the REST api GET /monitoringrules/{id} that returns the JSON representation of such rules.

### DIFF
--- a/planner/core/src/main/java/eu/seaclouds/platform/planner/core/DamGenerator.java
+++ b/planner/core/src/main/java/eu/seaclouds/platform/planner/core/DamGenerator.java
@@ -77,7 +77,8 @@ public class DamGenerator {
     public static final String SEACLOUDS_NODE_PREFIX = "seaclouds.nodes";
 
     private static DeployerTypesResolver deployerTypesResolver;
-    static Map<String, MonitoringInfo> monitoringInfoByApplication=new HashMap<>();
+    public static Map<String, MonitoringInfo> monitoringInfoByApplication=new HashMap<>();
+
 
     public static String generateDam(String adp, String monitorGenURL, String monitorGenPort, String slaGenURL, String influxdbURL, String influxdbPort){
         Yaml yml = new Yaml();
--- a/planner/service/src/main/java/eu/seaclouds/platform/planner/service/RulesResource.java
+++ b/planner/service/src/main/java/eu/seaclouds/platform/planner/service/RulesResource.java
@@ -1,0 +1,44 @@
+package eu.seaclouds.platform.planner.service;
+
+import com.codahale.metrics.annotation.Timed;
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import eu.seaclouds.monitor.monitoringdamgenerator.MonitoringInfo;
+import eu.seaclouds.platform.planner.core.DamGenerator;
+import it.polimi.tower4clouds.rules.MonitoringRules;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import javax.ws.rs.*;
+import javax.ws.rs.core.MediaType;
+
+/**
+ * Copyright 2014 SeaClouds
+ * Contact: SeaClouds
+ * <p>
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * <p>
+ * http://www.apache.org/licenses/LICENSE-2.0
+ * <p>
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+@Path("/monitoringrules")
+@Produces(MediaType.APPLICATION_XML)
+public class RulesResource {
+
+    public RulesResource() {}
+
+    @GET
+    @Path("/{id}")
+    @Timed
+    public MonitoringRules getMonitoingRules(@PathParam("id") String id){
+        return DamGenerator.monitoringInfoByApplication.get(id).getApplicationMonitoringRules();
+    }
+}
--- a/planner/service/src/main/java/eu/seaclouds/platform/planner/service/PlannerApplication.java
+++ b/planner/service/src/main/java/eu/seaclouds/platform/planner/service/PlannerApplication.java
@@ -35,10 +35,12 @@ public class PlannerApplication extends Application<PlannerConfiguration> {
                                                 plannerConfiguration.getSlaGeneratorURL(),
                                                 plannerConfiguration.getInfluxdbURL(),
                                                 plannerConfiguration.getInfluxdbPort());
+        RulesResource rr = new RulesResource();
 
         environment.jersey().register(pr);
         environment.jersey().register(rpr);
         environment.jersey().register(dgr);
+        environment.jersey().register(rr);
 
         //TODO: add health ckecks
     }


planner/service/src/main/java/eu/seaclouds/platform/planner/service/RulesResource.java
```diff
@@ -0,0 +1,58 @@
...
+ * <p>
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+@Path("/monitoringrules")
+@Produces(MediaType.APPLICATION_JSON)
```
andreaturli: why this annotation if the Resource is not producing any json ?

planner/service/src/main/java/eu/seaclouds/platform/planner/service/RulesResource.java
```diff
@@ -0,0 +1,58 @@
...
+    static Logger log = LoggerFactory.getLogger(RulesResource.class);
+
+    public RulesResource() {
+
+    }
+
+    @GET
+    @Path("/{id}")
+    @Timed
+    public Object damGenPost(@PathParam("id") String id){
```
andreaturli: is this method generating a dam?

andreaturli: I'm quite sure you can write something like the following:
```
    @GET
    @Timed
    public MonitoringInfo getMonitoringInfo(@ PathParam("id") String id) {
        return DamGenerator.monitoringInfoByApplication.get(id);
    }
```
isn't that easier?

szenzaro: you are right, copy and paste mistake :p

planner/service/src/main/java/eu/seaclouds/platform/planner/service/RulesResource.java
szenzaro: Sorry I do not understand your comment. The output of the /monitoringrules/{id} API is a JSON String

planner/service/src/main/java/eu/seaclouds/platform/planner/service/RulesResource.java
andreaturli: you declared "@Produces(MediaType.APPLICATION_JSON)" so why XML?

adriannieto:
@szenzaro @andreaturli Dashboard will translate the response into an object in order avoid boilerplate code[1]. As part of my work on the dashboard I'm going to work in the integration of the Java clients of the other components, like SLA and T4C (also Brooklyn if I figure out a way to deal with the collisions between maven dependencies) so object as response (or at least compatible output in XML, because I don't know if T4C supports JSON, the code looks to be prepared for XML).

I think that you could talk with @MicheleGuerriero if you want an example of how to return monitoring rules directly. Be aware that if you directly import the model from Tower4Clouds Dropwizard will find some offending dependencies, please take a look to my work on [2]

I hope it will help, it's mandatory to move to objects (at least if they already exist), IMO is way more important than deadlines to have good code. 

 [1] https://github.com/adriannieto/SeaCloudsPlatform/blob/feature/move-orchestration-from-js-to-java/dashboard/src/main/java/eu/seaclouds/platform/dashboard/rest/DeployerResource.java#L76 

[2] https://github.com/adriannieto/SeaCloudsPlatform/blob/feature/move-orchestration-from-js-to-java/dashboard/pom.xml#L74

My two cents.

szenzaro:
Ok so I will go for Andrea's suggestion. Returning the object will result in returning an xml that with your code will be translated to the MonitoringRules java object

szenzaro:
Ok, I refactored it to be compliant with your comment.

szenzaro:
I refactored and rebased it. Is it better now?

andreaturli:
@szenzaro, can you please add a test. I think http://www.dropwizard.io/0.9.1/docs/manual/testing.html#testing-resources
may be the right approach. wdyt?

planner/service/src/main/java/eu/seaclouds/platform/planner/service/RulesResource.java
```diff
@@ -0,0 +1,48 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+@Path("/monitoringrules")
+public class RulesResource {
+
+    static Logger log = LoggerFactory.getLogger(RulesResource.class);
+
+    public RulesResource() {
```
andreaturli: is it useful?

planner/service/src/main/java/eu/seaclouds/platform/planner/service/RulesResource.java
```diff
@@ -0,0 +1,48 @@
...
+ * http://www.apache.org/licenses/LICENSE-2.0
+ * <p>
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+@Path("/monitoringrules")
```
andreaturli: isn't `@Produces(MediaType.APPLICATION_JSON)` needed, as suggested [here](http://www.dropwizard.io/0.9.1/docs/getting-started.html#creating-a-resource-class) ?

planner/service/src/main/java/eu/seaclouds/platform/planner/service/RulesResource.java
szenzaro: Probably now it can be removed. I will remove it.

planner/service/src/main/java/eu/seaclouds/platform/planner/service/RulesResource.java
szenzaro: I added MediaType.APPLICATION_XML since Monitoring rulesed are serialized to xml. thanks

jacopogiallo:
@andreaturli What's wrong with this PR? I see a error only in the DAMGenerator class, which is however outdated wrt the new version developed by @kiuby88 

No criticism, I am just trying to understand :)

adriannieto:
@szenzaro could you update fix the building error? It seems that this PR is out of sync with the latest changes as @jacopogiallo said.

szenzaro:
@adriannieto @jacopogiallo could you please check if it is ok now?
If yes I will squash the commits.

planner/core/src/main/java/eu/seaclouds/platform/planner/core/DamGenerator.java
```diff
@@ -77,7 +77,7 @@
     public static final String SEACLOUDS_NODE_PREFIX = "seaclouds.nodes";
 
     private static DeployerTypesResolver deployerTypesResolver;
-    static Map<String, MonitoringInfo> monitoringInfoByApplication=new HashMap<>();
+    public static Map<String, MonitoringInfo> monitoringInfoByApplication=new HashMap<>();
```
adriannieto: Are you sure that public visibility is mandatory?

adriannieto:
As @szenzaro stopped working for SeaClouds, I'd try to simplify the review allowing it to be merged without tests, but @jacopogiallo should be aware that probably this could lead into troubles later.

jacopogiallo:
Thanks @adriannieto.

szenzaro:
@adriannieto the situation is not that simple (I did not just "stopped" to work on SeaClouds), but thank you for understanding priorities.

adriannieto:
Yes, rebase it, and I will take a final look to the code.

szenzaro:
@adriannieto I rebased the PR, could you please have the final look at it?

adriannieto:
@MicheleGuerriero do you think this is ready to merge? (I'm asking you because you're related with the Monitoring)

MicheleGuerriero:
@adriannieto the code implementing the integration between the DAMGenerator and the monitoring-dam-generator is in planner/core and since this PR is just providing a service for returning the monitoring rules of an application that integration is still the one on the master. I cannot add much more about the design of the service since I was not involved in that. The service looks like it is returning a list of monitoring rules. If it is designed as what you expect from the Dashboard and you, as original the reviewer, think the code is fine, I think it could be merged.

jacopogiallo:
@adriannieto @MicheleGuerriero What about this PR? Is it ready to be merged? I see it has again conflicts with the rest of the SeaClouds platform (even if it has already been rebased). In case this PR is ready to be merged, I can @szenzaro whether he can can kindly rebase it again (as you know, he is not working for SeaClouds any more - and I have to ask it to him as a personal favor).

FYI, this PR is almost three months old.

szenzaro:
@jacopogiallo I rebased as you asked ;)

jacopogiallo:
Thanks @szenzaro :)

@MicheleGuerriero @adriannieto Can you complete reviewing or merge the PR, please?

jacopogiallo:
@/all Can anybody perform this review, please?

PR #196, opened 2015-12-16
Updating Normative Types version that is used by AAM

kiuby88:


andreaturli:
Thanks @kiuby88 
do we need to update also the version used by a4c's ToscaParser ?

kiuby88:
Thanks for your reply, @andreaturli.
It is not necessary. a4c's ToscaParser is already using this version.
But `tosca-normative-types:1.0.0.wd03-SNAPSHOT` is present in a lot of file still.

andreaturli:
thanks @kiuby88

kiuby88:
Thx @andreaturli

PR #197, opened 2015-12-16
Fix/dam metadata

kiuby88:
Adding expected metadata to DAM.


andreaturli:
thanks @kiuby88

kiuby88:
Thx @andreaturli

PR #203, opened 2015-12-18
FIX: naming autoscaling policy type

perezp:
This PR addresses a bug on the name of the type of autoscaling policies. 
@kiuby88 Can you take al look?


kiuby88:
Thanks a lot @perezp 
+1

PR #206, opened 2016-01-11
monitoring-dam-generator minor fixes plus InfluxDB integration at design time for metric visualization.

MicheleGuerriero:
This PR apply some fixes to the monitoring-dam-generator-core and also wants to add design time support for InfluxDB, in oder to have it properly configured for the metric visualization at runtime.

In particular the PR carries the following additions:

1) env->shell.env in the data collectors node template, according to the provided [specification](https://docs.google.com/document/d/1Z9RdaWn54IomgI-jCiw-tdutsCdDoHfk67PNUASOFmY/edit?disco=AAAAAjfm8b4)

2) env vars from List<Map> to Map in the data collectors node template according to the provided [specification](https://docs.google.com/document/d/1Z9RdaWn54IomgI-jCiw-tdutsCdDoHfk67PNUASOFmY/edit?disco=AAAAAjfm8b4)

3) now the DataCollector node type is also inserted by the monitoring-dam-generator in the output ADP.

4) added data collectors "start.sh" scripts in the resources folder.

5) monitoring-dam-generator-core modified to support the integration with InfluxDB. In particular the data collector node template is configured with information about InfluxDB IP and port.

6) planner-core and planner service modified to manage infomation about InfluxDB endpoint and pass them to the monitoring-dam-generator-core. In particular InfluxDB IP and port are taken from the config.yaml file of planner-service.

7) monitoring-dam-generator-service modified to configure its core with InfluxDB endpoint information.
(this point is just for maintenance since this component is not used anymore). 

@szenzaro and @kiuby88 can you please review this PR? In particular I think point 1), 2), 3) and 4) could be addressed by @kiuby88 while 6) and 7) by @szenzaro since they focus on the Planner. Thanks in advance.


monitor/monitoring-dam-generator-core/src/main/java/eu/seaclouds/monitor/monitoringdamgenerator/dcgenerators/MODACloudsDcGenerator.java
```diff
@@ -17,10 +15,10 @@
...
 
     private static final String MODACLOUDS_DC_ID = "modacloudsDc";
-    private static final String START_SCRIPT_URL = "http://start.sh/";
+    private static final String START_SCRIPT_URL = "http://installMODACloudsDc.sh/";
 
     public void addDataCollector(Module module, String monitoringManagerIp,
-            int monitoringManagerPort) {
+            int monitoringManagerPort, String influxdbIp, int influxdbPort) {
 
         if(module.getHost().getDeploymentType().equals("IaaS")){
```
adriannieto: IaaS could be part of an enum WDYT?

monitor/monitoring-dam-generator-core/src/test/java/eu/seaclouds/monitor/monitoringdamgenerator/tests/MonitoringGenerationTest.java
```diff
@@ -28,7 +28,7 @@ public void monitoringInformationGenerationTest() throws Exception {
 
         Map<String, String> parametersTest;
         
-        MonitoringDamGenerator service = new MonitoringDamGenerator(new URL("http://127.0.0.1:8170"));
+        MonitoringDamGenerator service = new MonitoringDamGenerator(new URL("http://127.0.0.1:8170"), new URL("http://127.0.0.1:8083"));
```
adriannieto: Are you starting two servers before the tests? Are they mocked or sth like this? Just to clarify, could you briefly explain me how this service is started with the local ips?

adriannieto:
One thing regarding to the tests, you don't have to define a test package inside the package you want to test.

For example, right now you have ```...test/java/eu/seaclouds/monitor/monitoringdamgenerator/tests/MonitoringGenerationTest.java```
 and this seems to be an incorrect usage of the packages.

If you are testing classes within eu.seaclouds.monitor.monitoringdamgenerator the the test shoud be in the same package so the path will become ```...test/java/eu/seaclouds/monitor/monitoringdamgenerator/MonitoringGenerationTest.java```  Take into account that this allows you to define helpful package methods for the tests in the main class that are not accessible from other packages.

monitor/monitoring-dam-generator-core/src/main/java/eu/seaclouds/monitor/monitoringdamgenerator/dcgenerators/MODACloudsDcGenerator.java
```diff
@@ -17,10 +15,10 @@
             .getLogger(MODACloudsDcGenerator.class);
 
     private static final String MODACLOUDS_DC_ID = "modacloudsDc";
-    private static final String START_SCRIPT_URL = "http://start.sh/";
+    private static final String START_SCRIPT_URL = "http://installMODACloudsDc.sh/";
```
adriannieto: Http? why not file or even relative path "start.sh"?

monitor/monitoring-dam-generator-service/pom.xml
```diff
@@ -1,77 +1,78 @@
...
+		<groupId>eu.seaclouds-project.monitor</groupId>
+		<artifactId>monitor</artifactId>
+		<version>0.8.0-SNAPSHOT</version>
+	</parent>
+	<artifactId>monitoring-dam-generator-service</artifactId>
+	<name>monitoring-dam-generator-service</name>
+	<url>http://maven.apache.org</url>
+	<properties>
+		<project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
+		<main.class>eu.seaclouds.monitor.monitoringdamgeneratorservice.ServiceLauncher</main.class>
```
adriannieto: Why are you using this?

monitor/monitoring-dam-generator-core/src/test/java/eu/seaclouds/monitor/monitoringdamgenerator/tests/MonitoringGenerationTest.java
MicheleGuerriero: @adriannieto Any server is started since any communication is realized. The url is taken just as input to properly configure data collectors with the IP and port of T4C, but they will be the DCs themselves to contact T4C sending their configuration after the deployment.

monitor/monitoring-dam-generator-core/src/main/java/eu/seaclouds/monitor/monitoringdamgenerator/dcgenerators/MODACloudsDcGenerator.java
MicheleGuerriero: This URL will be updated as soon as this PR will be accepted since they need to point to the URL of the two scripts under the resources folder. So when this scripts will be updated somewhere (in our master according to this PR), those URLs need to be updated with a next PR. In any case they need to be URLs and not paths, I think @kiuby88 can be more precise on this last point.

monitor/monitoring-dam-generator-service/pom.xml
MicheleGuerriero: Do you mean the main.class tag? It is simply recalled in the maven-shade plugin to point to the right main. The class eu.seaclouds.monitor.monitoringdamgeneratorservice.ServiceLauncher is the application starter.

monitor/monitoring-dam-generator-core/src/main/java/eu/seaclouds/monitor/monitoringdamgenerator/dcgenerators/MODACloudsDcGenerator.java
MicheleGuerriero: Sure, I will update this. Thanks!

MicheleGuerriero:
@adriannieto Thanks for the suggestion about the management of the test package. I will employ the practice you suggest updating the project structure. Thanks.

monitor/monitoring-dam-generator-core/src/main/java/eu/seaclouds/monitor/monitoringdamgenerator/dcgenerators/MODACloudsDcGenerator.java
kiuby88: Yes, we need to publish these files. For example, we could create a new repository in SeaClouds organization which contains the required files, upload these files to SeaCloudsPlatform repository, Amazon S3...

MicheleGuerriero: Is the current solution I described before compliant with the first one you propose or it needs to be implemented differently?

adriannieto: I would upload to Amazon or something like this to have before accepting the PR the resource url ready, but pointing to a github file in a quick PR will be accepted is fine also for me

MicheleGuerriero: Uploaded the script on S3. Thanks!

adriannieto:
Thank you for your answers, most of them were just because I was curious. I will continue reviewing it after the test package refactoring.

MicheleGuerriero:
I should have fixed the issue on the test package structure. @adriannieto can you check this?

monitor/monitoring-dam-generator-core/src/main/java/eu/seaclouds/monitor/monitoringdamgenerator/dcgenerators/MODACloudsDcGenerator.java
adriannieto: Use enums ASAP.

adriannieto:
I think it could be merged. Rebase and merge it.

MicheleGuerriero:
@adriannieto thanks a lot for the review! I rebased and squashed the commits.

adriannieto:
LGTM. +1

adriannieto:
I found an issue, although you define InfluxDB endpoint for DropWizard, I realized that you didn't update the catalog to add the necessary wiring.  

https://github.com/SeaCloudsEU/SeaCloudsPlatform/pull/206/files#diff-6cc68a0e440463577d5f6c5b182f0624R43

So if you try to deploy this with SeaClouds it will fail by not allowing you to run the dropwizard jar.

MicheleGuerriero:
Yes you are right, actually I'm aware about it and it is related to the issue of having influxdb and griffin integrated in the catalog. Of course this is anticipating something, but I think it should work when we will update the catalog, am I right? Do you suggest to first update the catalog and then merge this PR?

adriannieto:
I think yes, but it's better to update the catalog and then merge this. I dunno why but grafana s failing to install by any particular reason in this branch https://github.com/adriannieto/SeaCloudsPlatform/tree/feature/grafana-integration maybe @andreaturli can help us on this.

adriannieto:
@MicheleGuerriero I reconsidered my proposal, let's hold the integration within the catalog until we have something working. But take into account that you will have to provide probably an image of a functional version of this when we will do the demo.  

LGTM

PR #207, opened 2016-01-12
Adding first brooklyn-tosca support.

kiuby88:
Adding first brooklyn TOSCA support to SeaClouds Deployer.
Next step would be to use maven instead of a fixed jar.


andreaturli:
thanks, @kiuby88

kiuby88:
You're welcome, @andreaturli

PR #208, opened 2016-01-12
FEATURE(Planner) Refactor of plan() method that now takes an AAM as argument. Added a test for the plan() method.

PaoloCifariello:
The partial refactor does not change the logic behind the Planner, but it makes the code of `plan()` more clear and intuitive.
I added a test for the `plan()` method that checks the information about deployment in the generated plans.
Could you please check if the ADP used inside the test is valid? (`planner/core/src/test/resources/adp_atos_case_study10Dec.yml`)

### DIFF
--- a/planner/core/src/main/java/eu/seaclouds/platform/planner/core/Planner.java
+++ b/planner/core/src/main/java/eu/seaclouds/platform/planner/core/Planner.java
@@ -23,7 +23,6 @@ import alien4cloud.tosca.parser.ParsingResult;
 import com.fasterxml.jackson.annotation.JsonProperty;
 import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.databind.ObjectMapper;
-import com.google.common.io.Resources;
 import eu.seaclouds.common.tosca.ToscaSerializer;
 import eu.seaclouds.planner.matchmaker.Matchmaker;
 import eu.seaclouds.planner.matchmaker.Pair;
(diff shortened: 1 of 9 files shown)


planner/core/src/test/java/eu/seaclouds/platform/planner/core/PlannerTest.java
```diff
@@ -0,0 +1,90 @@
...
+ * http://www.apache.org/licenses/LICENSE-2.0
+ * <p>
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+@Test
+public class PlannerTest {
```
andreaturli: How much different is this `PlannerTest` from [OptmizerTest](https://github.com/SeaCloudsEU/SeaCloudsPlatform/blob/master/planner/optimizer/optimizer-core/src/test/java/eu/seaclouds/platform/planner/optimizerTest/OptimizerTest.java) ?

Frankly speaking, I still don't see the reason of having a planner which is trivially a wrapper of the optimizer.

@perezp am I missing something?

jacopogiallo: The planner is also matchmaking technical requirements (which is something not performed by the Optimizer). If you like, we can extend the list of offerings with some offerings that are not matching tech reqs. 

Actually, there is also the discovery step (which we are omitting from testing, since we want a are testing with a static list of offerings, as you kindly suggested).

jacopogiallo: > Frankly speaking, I still don't see the reason of having a planner which is trivially a wrapper of the optimizer.

@andreaturli Actually, this comment seems to be more at the level of the architecture, rather than at the level of the code. I don't know whether it is appropriate to do it here.

javicubo:
@andreaturli The planner is not a wrapper of the optimizer. The optimizer simply makes an optimization of the previous steps where the Matchmaker calls the Discoverer. Also the Planner has as final output the DAM, which is not the output of the Optimizer, but also another action inside the Planner, the DAM generator.

planner/core/src/main/java/eu/seaclouds/platform/planner/core/Planner.java
```diff
@@ -44,31 +41,22 @@
     private static final String DISCOVERER_PATH = "discoverer/";
```
perezp: Planner is writing logs as ToscaSerializer.class

perezp:
@PaoloCifariello  for me the code is OK. 
I think it is too verbose in its normal operation, but I do not propose nor ask for any change of it, beyond the TOSCAserializer.class I commented before. 

I have had some issues while testing this PR. Also travis seems to show different types of outputs now ( https://travis-ci.org/SeaCloudsEU/SeaCloudsPlatform/builds/101885100 )

Can you paste in gitter the output that you see in the log with "mvn clean install" of the planner-core and I'll review it?

PaoloCifariello:
Thanks @perezp. I have solved this issue solved (which was a dependency problem) by adding the necessary exclusions to the pom.xml and by properly fixing configuration in log4j.xml

adriannieto:
@PaoloCifariello Nice catch, this project requires a HUGE clean-up to unify all the dependencies but, I think it's a little bit late now. I'm glad to know that you survived to Maven dependency hell.

PaoloCifariello:
@adriannieto Well, I have to thank you for your suggestion about using `mvn dependency:tree`, it was so helpful :smile:

PaoloCifariello:
@perezp @adriannieto Thank you for your help. Did i miss something? If not, can you please merge this PR?

perezp:
@PaoloCifariello 
Can you squash the commits and I'll try again to build and test?

perezp:
Thank you @PaoloCifariello . I've executed the tests and I liked to see that they produce the expected results. 

Now travis passes, but there are still generated a lot of "Debug" and "Trace" messages from Alien4cloud; for example, from these classes: 
DefaultListableBeanFactory
CommonAnnotationBeanPostProcessor
ClassPathBeanDefinitionScanner
AbstractMapper
CachedIntrospectionResults
InjectionMetadata
StandardEnvironment
ConfigurationClassBeanDefinitionReader
PlatformResourceBundleLocator

At present, the size log created from the testing of the planner-core module is 512K . 
For me, it's already +1 because the PR already adds value to the current version and it's not making the build fail. Nevertheless, for the next PR, please, consider reducing the log level of alien4cloud.

PR #209, opened 2016-01-18
FEATURE(Discoverer): introduction of MongoDB, added documentation, added new "fetchif" API to get filtered offerings

PaoloCifariello:
- The Discoverer uses a MongoDB service to store offerings, instead of using folders on the file system
- I have also added the documentation and some tests
- I have added a new "fetchif" API to get offerings filtered by properties (constraints are expressed in JSON): 
e.g. sending an HTTP request to `http://discovererurl/fetchif?constraints={"num_cpus":"8"}` will return a list of offerings about services with 8 cpus


adriannieto:
Mongodb is a great idea, but It involves the usage of new services, WDYT about Mapdb which is a Java based storage? Dashboard PR involves this in SeaCloudsApplicationDataStorage.java

jacopogiallo:
Hi @adriannieto. We decided to use a service like MongoDB since the consortium required a persistent DB that can be accessed by multiple instances of the Discoverer in a consistent manner (for replication purposes). A DB like MapDB would not satisfy this requirement. :)

adriannieto:
@jacopogiallo Sorry I wasn't aware about this, then you should add if MongoDB requires to be deployed to run the discoverer it should be part of the catalog.bom

jacopogiallo:
@adriannieto Yes, indeed @PaoloCifariello has added this dependency information in the "Requirements" section of the "README.md" file (https://github.com/PaoloCifariello/SeaCloudsPlatform/blob/discoverer-mongodb/discoverer/README.md#requirements)

adriannieto:
@andreaturli should we add MongoDB to catalog.bom?

jacopogiallo:
Yes, we will update it, if necessary. :)

@adriannieto, @andreaturli AFAIK there is also a starting image to load on the Discoverer's MongoDB. @PaoloCifariello will provide it.

adriannieto:
@jacopogiallo  we could work later in the catalog if you want. We could work with preloaded instance if necessary and then after everything is finished we can update the catalog.

jacopogiallo:
@adriannieto That's precisely what we (@PaoloCifariello and myself) were thinking to do, at least for the demo we are going to perform next thursday.

andreaturli:
@adriannieto the `catalog.bom` will be updated once all the components will be stable enough. 

@jacopogiallo be sure that the `installation guide` of the discoverer is clear enough. Ideally something like [this](https://brooklyn.apache.org/v/latest/start/running.html) should be provided for all of the components: a very simple set of steps to install and run the component, I think.

discoverer/src/main/java/eu/seaclouds/platform/discoverer/api/FetchIfAPI.java
```diff
@@ -0,0 +1,128 @@
...
+            String constraintValue = (String) constraintsObject.get(key);
+
+            if (!satisfyConstraint(constraintName, constraintValue, toscaString))
+                return false;
+        }
+
+        return true;
+    }
+
+    private boolean satisfyConstraint(String constraintName, String constraintValue, String toscaString) {
```
adriannieto: This is a great example of why using strings instead of objects is a really bad practice. This code is almost imposible to understand if you doesn't know TOSCA syntax, it will break silently if the syntax changes and it even hard to write tests for this. 

I'm not telling you to change this method, because SeaClouds Consortium already refused the usage of a shared object model because it wasn't a priority. I just want to expose the big issue that we have here in terms of reusability of the code.

jacopogiallo:
@andreaturli The README.md developed by @PaoloCifariello should be in line with those developed for the other SeaClouds' components. If you think this is not enough, perhaps we should extend this issue to the whole platform, not only for the Discoverer.

As always, no criticism. I'm just trying to be constructive. :)

discoverer/src/main/java/eu/seaclouds/platform/discoverer/api/FetchIfAPI.java
jacopogiallo: Yes, I agree with you @adriannieto. This was a decision the whole consortium took one year ago during weekly integration meetings, which hampers reusability of the code.

andreaturli: @adriannieto this is not a good example for the problem as it is a private method: the real problem is on the public interface of the services.

Also, nobody is asking to create a new object model as the tosca parser borrowed from a4c already gives us the pojos to work with. It is just a matter of using them, I guess.

andreaturli:
@adriannieto have you done with the review? do you think it is ok to merge ?

discoverer/README.md
```diff
@@ -0,0 +1,38 @@
...
+The SeaClouds Discoverer module provides information about cloud offering used by SeaCloudsPlatform.
+
+# Installing and running SeaClouds Discoverer
+The easiest way to install and run this component is explained in SeaClouds [README.md](../README.md).
+
+## Requirements
+Running SeaClouds Discoverer only requires Java 7 (or greater) installed on the target machine. Among with the software requirements, it also needs:
+
+- A discovererconf.yml configuration file.
+- A running MongoDB service
```
andreaturli: can you please clarify how the mongoDB will be used?

discoverer/README.md
```diff
@@ -0,0 +1,38 @@
+SeaClouds Discoverer
+====================
+
+The SeaClouds Discoverer module provides information about cloud offering used by SeaCloudsPlatform.
+
```
andreaturli: - how?
- what kind of information? 
- where are those information fetched from? 
- why? what problem does it solve?
- and so on ...

discoverer/src/main/java/eu/seaclouds/platform/discoverer/api/FetchIfAPI.java
```diff
@@ -0,0 +1,128 @@
...
+    private Discoverer discoverer;
+    private JSONParser JSONParser;
+    static Logger log = LoggerFactory.getLogger(FetchIfAPI.class);
+
+    public FetchIfAPI(Discoverer discoverer) {
+        this.discoverer = discoverer;
+        this.JSONParser = new JSONParser();
+    }
+
+    @POST
```
adriannieto: POST + getOfferingsIf doesn't make sense. IMO it should be GET

discoverer/src/main/java/eu/seaclouds/platform/discoverer/service/DiscovererApplication.java
```diff
@@ -0,0 +1,30 @@
...
+        MongoClient mongoClient = new MongoClient(configuration.getDatabaseURL(), configuration.getDatabasePort());
+
+        Discoverer discoverer = new Discoverer(mongoClient, configuration.getActiveCrawlers());
+
+        environment.jersey().register(new FetchAPI(discoverer));          /* /fetchOffer */
+        environment.jersey().register(new FetchAllAPI(discoverer));       /* /fetch_all */
+        environment.jersey().register(new DeleteAPI(discoverer));         /* /delete */
+        environment.jersey().register(new StatisticsAPI(discoverer));     /* /statistics */
+        environment.jersey().register(new RefreshAPI(discoverer));        /* /refresh */
+        environment.jersey().register(new FetchIfAPI(discoverer));
```
andreaturli: what's the advantage of having so many `*API` ? Couldn't they be collapsed into `DiscovererApi` maybe?

discoverer/src/main/java/eu/seaclouds/platform/discoverer/service/DiscovererApplication.java
```diff
@@ -0,0 +1,30 @@
...
+    public static void main(String[] args) throws Exception {
+        new DiscovererApplication().run(args);
+    }
+
+    @Override
+    public void run(DiscovererConfiguration configuration, Environment environment) {
+
+        MongoClient mongoClient = new MongoClient(configuration.getDatabaseURL(), configuration.getDatabasePort());
+
+        Discoverer discoverer = new Discoverer(mongoClient, configuration.getActiveCrawlers());
```
andreaturli: why the `Discoverer` object needs to be exposed here? Wouldn't be enough to pass the `mongoClient` to the `*API` ?

discoverer/src/main/java/eu/seaclouds/platform/discoverer/core/Discoverer.java
```diff
@@ -17,39 +17,51 @@
...
-        this.offeringManager = new OfferingManager();
+        this.activeCrawlers = activeCrawlers;
     }
 
+    public Discoverer(MongoClient mongoClient) {
+        this(mongoClient, new ArrayList<String>());
+    }
+
+    /* vars */
+    public OfferingManager offeringManager;
```
andreaturli: What is the role of this `OfferingManager`? is it a [DAO](https://en.wikipedia.org/wiki/Data_access_object) maybe?

If so, mabe something similar to what it is suggested on Dropwizard website ( [JDBI](http://www.dropwizard.io/0.9.2/docs/manual/jdbi.html) or [Hibernate](http://www.dropwizard.io/0.9.2/docs/manual/hibernate.html)) may be what you want to use to redesign the Discoverer.

Thoughts?

andreaturli:
@PaoloCifariello and @jacopogiallo I'm really sorry but I can't understand the purpose of this PR.
First of all, it changes too many things at once, from the architecture of the component to very low-level coding details so it is hard to follow, i.e., I don't know why you want MongoDB vs MySql so I can't help with this decision.

IMHO, it is already too complex for me to understand what's the purpose of the Discoverer and the problem it is trying to solve: I think this implementation is completely different from what has been discussed by the consortium during the last 2 years of project.

I think a better documentation would certainly help whoever wants to review this PR.

@fradandria do you have thoughts on this?

jacopogiallo:
@andreaturli The introduction of a DB has been explicitly requested by @fradandria and by the consortium as a whole (during the meeting in Malaga). We worked on it, and found MongoDB as a solution since we wanted to store the offerings in their YAML format (thus we needed a NoSQL database).

We can however work on the documentation, and try to improve it. @PaoloCifariello is already working on it.

discoverer/README.md
jacopogiallo: @PaoloCifariello is working on the documentation to improve it (by answering to the questions you provided).

discoverer/README.md
jacopogiallo: @PaoloCifariello is currently working on the documentation, and the new version will explain the usage of MongoDB.

Again, the DB was requested by @fradandria in Malaga. We added it to ensure persistency and consistency of offerings, especially in the case that SeaClouds is deployed on a cloud platform and multiple instances of the Discoverer have to access the DB. Furthermore, by making a snapshot of the DB we can also migrate it on a different platforms.

We decided to use MongoDB, since we were left free in choosing the technology (also this was discussed in the meeting in Malaga).

discoverer/src/main/java/eu/seaclouds/platform/discoverer/api/FetchIfAPI.java
jacopogiallo: @PaoloCifariello I agree with @adriannieto. Could you please fix it?

discoverer/src/main/java/eu/seaclouds/platform/discoverer/service/DiscovererApplication.java
jacopogiallo: @andreaturli I agree with you that this can be an alternative, and more elegant, solution. Implementing such a solution would consist in refactoring the code of the Discoverer, thus stealing time to other issues having higher priority. Would it be ok if we perform such refactoring in a future PR?

discoverer/src/main/java/eu/seaclouds/platform/discoverer/api/FetchIfAPI.java
PaoloCifariello: I also agree, I am going to fix it, thank you @adriannieto

fradandria:
@PaoloCifariello @andreaturli @jacopogiallo Actually I never asked to set up a DB. I said that we should be able to reuse data if Cloudify and Paasify were not available. we always talked  about a "proxy" solution. I was pointing out the funcionality more than the technology  used to implement such fucionality

jacopogiallo:
Ok, probably there was a misunderstanding among us. However, @fradandria isn't a DB a way to have consistent offerings when Cloudify and Paasify are not available?

discoverer/src/main/java/eu/seaclouds/platform/discoverer/core/Discoverer.java
```diff
@@ -17,39 +17,51 @@
...
     public Date lastCrawl;
 
-    public static Discoverer instance() { return singleInstance; }
+    public Discoverer(MongoClient mongoClient, ArrayList<String> activeCrawlers) {
+        this.initializeResources();
 
-    /* vars */
-    public OfferingManager offeringManager;
+        MongoDatabase db = mongoClient.getDatabase(this.databaseName);
+        MongoCollection<Document> coll = db.getCollection(this.collectionName);
```
andreaturli: who is responsible to create this collection?

fradandria: I guess UPI, what's happen is Cloudify and Paasify are down?

PaoloCifariello: MongoDB creates a collection implicitly when the collection is first referenced in a command, as stated on https://docs.mongodb.org/manual/reference/method/db.createCollection/

andreaturli: ok thanks, but who will fill that collection?

PaoloCifariello: The logic is unchanged, crawlers add offerings into the collection (instead of creating files on file system) and, in case CH or PaaSify are not available, we can import an image of the collection.

discoverer/src/main/java/eu/seaclouds/platform/discoverer/api/DiscovererAPI.java
```diff
@@ -0,0 +1,312 @@
...
+    private org.json.simple.parser.JSONParser JSONParser;
+    static Logger log = LoggerFactory.getLogger(DiscovererAPI.class);
+
+    public DiscovererAPI(Discoverer discoverer) {
+        this.discoverer = discoverer;
+        this.JSONParser = new JSONParser();
+    }
+
+    @POST
+    @Path("/delete")
```
andreaturli: `delete` what?
Maybe better use @DELETE?

discoverer/src/main/java/eu/seaclouds/platform/discoverer/api/DiscovererAPI.java
```diff
@@ -0,0 +1,312 @@
...
+    @Produces(MediaType.APPLICATION_JSON)
+    public DeleteRepresentation deleteOfferingById(@QueryParam("oid") String offerId)
+            throws IOException {
+
+        ArrayList<String> removedOfferingIds = new ArrayList<>();
```
andreaturli: why a list? I'm expecting `deleteOfferingById` to delete a single offering, so it should return a boolean, I think, surely not a list of a single ID.

discoverer/src/main/java/eu/seaclouds/platform/discoverer/api/DiscovererAPI.java
```diff
@@ -0,0 +1,312 @@
...
+
+        /* if the offering id is valid and the discoverer is able to remove the offering associated
+         * then the offering id is added to the list of removed offering  */
+        if(this.discoverer.removeOffering(offerId)) {
+            removedOfferingIds.add(offerId);
+        }
+
+        return new DeleteRepresentation(removedOfferingIds);
```
andreaturli: I think you should use `javax.ws.rs.core.Response` instead of these `weird` pojos

discoverer/src/main/java/eu/seaclouds/platform/discoverer/api/DiscovererAPI.java
```diff
@@ -0,0 +1,312 @@
...
+
+        if(offering != null) {
+            offeringTOSCA = offering.toscaString;
+        }
+
+        if (offeringTOSCA.isEmpty()) {
+            ids.clear();
+        }
+
+        return new FetchAllRepresentation(ids, offeringTOSCA);
```
andreaturli: I really think it would be much better something like

```

  @GET
  public List< Offering> list() {
    List<Offering> offerings = Lists.newArrayList();
    // your logic
    return offerings;
  }
```

discoverer/src/main/java/eu/seaclouds/platform/discoverer/api/DiscovererAPI.java
```diff
@@ -0,0 +1,312 @@
...
+        Collection<String> offeringIds = discoverer.getAllOfferingIds();
+        for (String offeringId : offeringIds) {
+            ids.add(offeringId);
+        }
+
+        return ids;
+    }
+
+    @POST
+    @Path("/fetch")
```
andreaturli: I guess, something like this would be a little cleaner and easier to read, test and maintain
```
  @GET
  @Path("{id}")
  public Offering get(@PathParam("id") String id) {
         FindIterable<Document> cursor = this.offeringsCollection.find(query);
         Offering offering =  Offering.fromDB(cursor.first());
         if (offering == null) {
                throw new WebApplicationException(Response.Status.NOT_FOUND);
    }
    return offering;
  }
```
notice, I've not compiled it so you may need to adjust it

andreaturli:
@PaoloCifariello I've not finished the review, as it is too complicated as I suggested in one of my previous comments. Anyway, I think you get the general idea

discoverer/src/main/java/eu/seaclouds/platform/discoverer/api/DiscovererAPI.java
PaoloCifariello: seems great, doing this way I can also avoid some useless code, thank you @andreaturli

PaoloCifariello: Unfortunately this does not apply here because this method is used to get all the offerings as a unique TOSCA YAML file, but it is fine for "fetchif" which precisely returns a list of Offerings.

andreaturli: what is `fetchif` ?

PaoloCifariello: It is used to get offerings by specifying constraints on properties: https://github.com/SeaCloudsEU/SeaCloudsPlatform/pull/209/files#diff-dea25a5853aa6e36f06e2db30866d009R131

fradandria:
@andreaturli @PaoloCifariello @jacopogiallo ... So, what is the solution? if the simplest solution is the utilization of Mongo DB. Could  we think to implement it just to speed up the delivery?

discoverer/src/main/java/eu/seaclouds/platform/discoverer/api/DiscovererAPI.java
andreaturli: You mean something like this?
```
  @GET
  @Path("/offerings")
  public List<Offering> list(@QueryParam("constraints") String constraints) {
    List<Offering> offerings = Lists.newArrayList();
    // your logic
    return offerings;
  }
```

PaoloCifariello: Yes, precisely, but with @Path("/fetchif") or whatever it will be the path of that api.

andreaturli: Never seen a path like that but do whatever you like

jacopogiallo:
@fradandria MongoDB is already considered in this solution, which IS WORKING. Perhaps, we could keep it and focus on issues with higher priority. WDYT?

fradandria:
I definitely agree!

andreaturli:
@fradandria can you vote then for this PR?
Is it ok to merge?

fradandria:
@andreaturli @PaoloCifariello @jacopogiallo If the PR is not breaking Seaclouds AT ALL and we do not NEED EXTRA WORK on making this working, I'd vote for merging the PR and going back to the issue as soon as we can less work than now...Do u agree? do you plan to  have any issues in the close  future?

jacopogiallo:
@fradandria @andreaturli The exposed API is unchanged, so (it has no conflicts with the Planner invoking it and) it won't break the SC Platform.

PR #213, opened 2016-01-21
Fix/aam relationships

rosogon:
This  PR defines relationships in a more-TOSCA way (still not TOSCA compliant: this needs also changes in the planner) and adds the needed changes in designer and aamwriter.

A relationship is now like:
```
requirements:
  - requirementName: targetNode
    type: relation_type
```

relation_type is one of the following: seaclouds.relations.databaseconnections.jdbc, seaclouds.relations.databaseconnections.php, seaclouds.relation.connection.endpoint.host

The designer now shows two more field in "edit link" form: 
* credentials_file: where to store the credentials in the deployed application
* type of relationship (relation_type above)

The aamwriter now writes the requirement in the form explained above and adds the credentials_file into the properties of the source node_template.

@kiuby88 : Can you review this?

### DIFF
--- a/planner/aamwriter/src/main/resources/examples/atos_case_study_aam.yml
+++ b/planner/aamwriter/src/main/resources/examples/atos_case_study_aam.yml
@@ -1,22 +1,28 @@
 tosca_definitions_version: tosca_simple_yaml_1_0_0_wd03
 description: Sample 3-tier application
 imports:
-- tosca-normative-types:1.0.0.wd03-SNAPSHOT
+- tosca-normative-types:1.0.0.wd06-SNAPSHOT
 topology_template:
   node_templates:
     www:
       type: sc_req.www
+      properties:
+        language: JAVA
       requirements:
       - endpoint: webservices
+        type: seaclouds.relation.connection.endpoint.host
     webservices:
       type: sc_req.webservices
       properties:
+        language: JAVA
+        location: ''
         java_version:
           constraints:
           - greater_or_equal: '7'
           - less_or_equal: '8'
       requirements:
       - endpoint: db1
+        type: seaclouds.relations.databaseconnections.jdbc
     db1:
       type: sc_req.db1
       properties:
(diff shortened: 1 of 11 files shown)


planner/aamwriter/src/main/resources/examples/atos_case_study_aam.yml
```diff
@@ -73,7 +73,7 @@ groups:
         benchmark_platform: hp_cloud_services.2xl
     - dependencies:
         operation_webservices: '2'
-    - QoSRequirements:
+    - AppQoSRequirements:
```
perezp: @rosogon , 
I see that this line is the only difference with respect the previous AAM. Can you provide us with the example of the aam generated now? The one with the requirements part changed. 

It's to use it as input in the Planner component and its modules.

rosogon: You are right. I will update the aams in the examples by tomorrow. Meanwhile, you check a generated aam here: https://gist.github.com/rosogon/d17a1196ca93bb58f599

rosogon:
Hi @perezp. I updated the JSONs and the AAMs. If you agree, I'll squash and merge. 

I think @MicheleGuerriero is also interested in the AAM changes.

adriannieto:
I think #215 is related with this

perezp:
@rosogon 
in my opinion the PR is OK. It improves the AAMs currently generated and can be merged. 

However, before merging, I would wait for comments from the consumers of the new information generated in the AAM, to check whether it is what they needed.

rosogon:
The consumers are @perezp (optimizer) and @kiuby88 (dam generator).

kiuby88:
Sorry @rosogon, I will review ASAP

kiuby88:
it is ok, for now. Probably, I think it may be necessary to modify the MySQL and PHP properties but we will need discuss about this. So, this first step is ok :+1:

rosogon:
Thanks for the responses. Merging

PR #214, opened 2016-01-22
Refactoring of the nuro DC toward the seaclouds-dc

MicheleGuerriero:
This PR modify what previously was the nuro-data-collector, that now is called seaclouds-data-collector. The data collector had some refactoring. Mainly:

- it now has a dropwizard service in order to be able to dynamically add new monitored resources at runtime.
- it is now also able to provide the availability of PaaS module, configuring the DC with the module URL by mean of a monitoring rule. After a new resource has been added and a monitoring rule asking for the availability of that resource has been installed into T4C, this DC will periodically ping the above url, returning 1 if it is reachable, 0 if not.
- something changed in how the DC is configured being now configurable partially at runtime. The installation details are reported in the README

@adriannieto can you have a look at this?


adriannieto:
This PR requires some work before tagging it as mergeable. First, try to reuse as much dependencies built in DW as possible, it will simplify a lot the detection of runtime dependency collisions. Also as you can see this PR decreases the coverage, please update the tests with the new changes.

MicheleGuerriero:
Thanks a lot @adriannieto! I will fix the issues in the .pom and add some tests (currently they are the same of the previous version and as you suggest they need to be improved).

monitor/seaclouds-data-collector/src/main/java/eu/seaclouds/monitor/seacloudsDc/Metric.java
```diff
@@ -93,29 +78,61 @@ public void update(Observable o, Object arg) {
...
     }
 
-    private void createTask(Timer timer, Resource resource, int samplingTime) {
-        timer.scheduleAtFixedRate(new MetricSender(resource), 0,
+    private void createTask(Timer timer, Resource resource, int samplingTime, String applicationUrl, String user, String password) {
+        timer.scheduleAtFixedRate(new MetricSender(resource, applicationUrl, user, password), 0,
                 samplingTime * 1000);
     }
 
     private class MetricSender extends TimerTask {
```
adriannieto: What I suggested was to extract this class to an standalone .java file. This could be done automatically. Maybe it's out of the scope of this PR but as you're refactoring the Data Collectors maybe you may consider it.

monitor/seaclouds-data-collector/pom.xml
```diff
@@ -0,0 +1,113 @@
...
+		</dependency>
+		
+		<dependency>
+	        <groupId>com.google.guava</groupId>
+            <artifactId>guava</artifactId>
+        </dependency>	
+
+		<dependency>
+			<groupId>org.testng</groupId>
+			<artifactId>testng</artifactId>
```
adriannieto: Please set scope to test

monitor/seaclouds-data-collector/src/main/java/eu/seaclouds/monitor/seacloudsDc/MainDc.java
```diff
@@ -0,0 +1,62 @@
...
+    public void initialize(Bootstrap<Configuration> bootstrap) {
+        bootstrap.setConfigurationSourceProvider(
+                new SubstitutingSourceProvider(bootstrap.getConfigurationSourceProvider(),
+                        new EnvironmentVariableSubstitutor(false)
+                )
+        );
+    }
+
+    @Override
+    public void run(Configuration configuration, Environment environment) throws Exception {
```
andreaturli: this is not the idiomatic way to initialize a dropwizard application.
please follow the dropwizard documentation, in particular the [getting started guide](http://www.dropwizard.io/0.9.2/docs/getting-started.html#creating-an-application-class): this will make things much easier to read

monitor/seaclouds-data-collector/src/main/java/eu/seaclouds/monitor/seacloudsDc/DcService.java
```diff
@@ -0,0 +1,45 @@
+package eu.seaclouds.monitor.seacloudsDc;
```
andreaturli: I think the right package should be `eu.project.seaclouds.monitor.datacollector`

monitor/seaclouds-data-collector/src/main/java/eu/seaclouds/monitor/seacloudsDc/Registry.java
```diff
@@ -1,16 +1,21 @@
...
 
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-import eu.seaclouds.monitor.nuroDc.metrics.NuroMetric;
+import eu.seaclouds.monitor.seacloudsDc.metrics.NuroMetric;
+import eu.seaclouds.monitor.seacloudsDc.metrics.ModuleAvailability;
 import it.polimi.tower4clouds.model.data_collectors.DCDescriptor;
 import it.polimi.tower4clouds.model.ontology.InternalComponent;
 import it.polimi.tower4clouds.model.ontology.Resource;
```
andreaturli: Could you please add some javadoc to this `Registry` class, its purpose is not clear.

andreaturli:
@MicheleGuerriero 
1. could you please explain why this datacollector needs to be a restful service?
2. What is the problem you are trying to solve? The implementation is quite convoluted and not immediately understandable IMHO.

monitor/seaclouds-data-collector/src/main/java/eu/seaclouds/monitor/seacloudsDc/DcService.java
```diff
@@ -0,0 +1,45 @@
...
+    
+    public DcService(Properties dcProp) {
+        
+        Registry.initialize(dcProp);
+
+    }
+    
+    @POST
+    @Path("/resource")
+    public Response addMonitoredResource(@QueryParam(TYPE_PARAMETER) String type, @QueryParam(ID_PARAMETER) String id,  @QueryParam(URL) String url) {
```
andreaturli: generally in a @POST you want to send a payload not query params. Any particular reason to not use a usual body?

MicheleGuerriero:
@andreaturli you are right, I know the way of works of this DC has significantly changed. The addressed problem is that in its previous version the DC was gathering the set of monitored resources from environmental variables in a static way. This means the DC was able to monitor just the initial set of resources according to its configuration. It was like this because it was assumed to work in a distributed way, in the sense that an instance of the DC was assigned to a specific VM/resource. If you have a cluster of VM instances each instance has an associated DC. Then T4C gather instances'data from each DC and is able to aggregate this data, for example averaging over the cluster. This was causing the DC to affect the application deployment (like currently the MODAClouds DC still does). This refactoring makes this DC centralized. This means basically that a single instance of the DC can monitor multiple instances, which are dynamically added. That's why I moved to a dropwizard application, just to expose a single endpoint to dynamically adds monitored resources during the runtime. In designing this solution I was first stimulated by the NURO case study which requires to not have the DCs affecting the application deployment, but then it became a more general solution. Actually I think it would be nice to have each data collector (where possible) to be centralized, but this is just my opinion and I could miss specific requirements that led the other DCs to be designed in such a distributed way. Does all of this makes sense? I hope I answered to your questions.

monitor/seaclouds-data-collector/src/main/java/eu/seaclouds/monitor/seacloudsDc/Metric.java
MicheleGuerriero: I'm still not completely convinced about the motivation behind this refactoring. My assumption was to have that class private within the abstract class because I don't want it to be available outside. Probably I'm wrong and I don't see your point, but I would not change this, at least without envisioning specific pros.

MicheleGuerriero:
@andreaturli @adriannieto thanks for your suggestions. I tried to address all your comments, do you think they are now fixed?

adriannieto:
@MicheleGuerriero  As this is required for the functional demo and we are running out of time, I think it looks good enough to merge it. Please squash first.

andreaturli:
@adriannieto what do you mean by good enough? This dc is replicating the expexted functionalities of the monitor server. Being late is not a good excuse to add tons of tech debt to this project, IMHO.

Anyway, merge that if you like.

andreaturli:
@adriannieto what do you mean by good enough? This dc is replicating the expected functionalities of the monitor server. Being late is not a good excuse to add tons of tech debt to this project, IMHO.

Anyway, merge that if you like.

adriannieto:
@andreaturli  sorry I didn't pay attention to the conversation, I just take a look to the code. About your discussion, in my opinion DW provides a  flexible way to enable and disable metrics on the fly. The monitor server be on charge to trigger this modifications, so no functionality will be replaced. 

IMO it seems nice to have a way to reconfigure the DC without deploying them again.

andreaturli:
Ever heard about YAGNI ?https://en.m.wikipedia.org/wiki/You_aren%27t_gonna_need_it

Anyway, feel free to merge it, mine it's just an opinion, I think.

perezp:
Thanks @adriannieto @MicheleGuerriero and @andreaturli

I've merged the PR but I cannot see the comment I wrote to justify the merge. I summarize: 
The conversation did not look to be solved timely. I would need more information to see the issue of the replication of functionalities. In my opinion, this PR added value to the current state, which was necessary.

@andreaturli I agree with a refactoring when everybody have clearer the replication of functionalities.

PR #215, opened 2016-01-23
FEATURE: utilization of benchmark platforms info in the Optimizer method

perezp:
This PR enhances the Optimizer capabilities to manage the information of the benchmark platforms (i.e., the cloud resources where application modules have been tested and measured). 

- It also adds to the pom.xml the control of licenses
- It adds license to the files which lacked it 
- It adds the names of the benchmark platforms to the GUI. 
- It solves bugs that have been found in the lasts days
- It contains tests of the management of the new information. 
- It contains example files used in the tests. 

@kiuby88 @adriannieto @andreaturli @rosogon  
Could you review the code and provide suggestions? 

@rosogon can you check whether the changes I made in dashboard/src/main/resources/webapp/static/lib/angular-topology-editor/editor.js are OK ? 

This PR looks long but differences in many of the files are just license addition and text formatting.


rosogon:
+1 on the javascript side

planner/optimizer/optimizer-core/src/main/java/eu/seaclouds/platform/planner/optimizer/Reoptimizer.java
```diff
@@ -54,7 +54,7 @@ public Reoptimizer(SearchMethodName name) {
       }
    }
 
-   public String[] optimize(String appModel, String suitableCloudOffer,
+   public String[] optimize(String appModel, String suitableCloudOffer, String benchmarkPlatforms,
          int numPlansToGenerate_ArrayLength, double hyst) {
       // TODO Auto-generated method stub
       return null;
```
adriannieto: @perezp If this functionality is not yet implemented why do you want to push it to the master? Maybe is better to leave in a separate branch or even in a stash until you develop it. In any case if the return type is an array, at least return an empty one to avoid NPE.

planner/optimizer/optimizer-core/src/main/java/eu/seaclouds/platform/planner/optimizer/util/YAMLmatchmakerToOptimizerParser.java
```diff
@@ -39,18 +40,18 @@
...
 
    static Logger log = LoggerFactory.getLogger(YAMLmatchmakerToOptimizerParser.class);
 
    @SuppressWarnings("unchecked")
    public static List<Object> getListofOptions(String appModel) {
       Yaml yamlApp = new Yaml();
-      if(log.isDebugEnabled()){
-      log.debug("Loading String to a YAML using sknakeyaml.Yaml");
+      if (log.isDebugEnabled()) {
+         log.debug("Loading String to a YAML using sknakeyaml.Yaml");
```
adriannieto: This doesn't make sense, log.debug() will be only printed to log stream if you're running with log level set to debug.

adriannieto:
Just as a suggestion, next time try to avoid to push totally unrelated things with the PR (like adding missing headers). You always can open two PR's at the same time. It's easier to review more focused PR's

planner/optimizer/optimizer-core/src/main/java/eu/seaclouds/platform/planner/optimizer/util/DefaultConstants.java
```diff
@@ -0,0 +1,24 @@
...
+ *    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *    See the License for the specific language governing permissions and
+ *    limitations under the License.
+ */
+
+package eu.seaclouds.platform.planner.optimizer.util;
+
+public class DefaultConstants {
+
+   public static final String DEFAULT_BENCHMARK_INFORMATION_PATH    = "./src/main/resources/benchmark-platforms.yml";
```
kiuby88: Probably, it may be better to find this file as a resource instead of use the path

planner/optimizer/optimizer-core/src/main/java/eu/seaclouds/platform/planner/optimizer/util/YAMLmatchmakerToOptimizerParser.java
perezp: Thank you @adriannieto  . The check improves slightly the performance because the string does not need to be created, the debug method invoked, and then ignored. It simply checks a condition and skips the rest of the process. 

OK, this string was not very complex to generate either...

planner/optimizer/optimizer-core/src/main/java/eu/seaclouds/platform/planner/optimizer/util/DefaultConstants.java
perezp: Thank you @kiuby88 . Suggestions on how to do it? 

Next line reads from the resources. 

I can be wrong, but what I've observed in the lasts days implementing and testing this feature is that: testing the feature with the packaged optimizer looks for it in the resources (and the next line gives just the name), while unit tests that have not packaged the code look for them in the path. 

If you give me any link or suggestion on how to implement it, I'm happy to make the modification.

planner/optimizer/optimizer-core/src/main/java/eu/seaclouds/platform/planner/optimizer/util/YAMLoptimizerParser.java
```diff
@@ -473,16 +475,26 @@ private static String getFinalHostNameOfModule(Map<String, Object> modules, Stri
...
          return topology;
       }
 
       TopologyElement newelement = new TopologyElement(elementName);
-      double hostPerformance = appInfoSuitableOptions.getCloudCharacteristics(elementName,
-            YAMLmodulesOptimizerParser.getMeasuredPerformanceHost(elementName, groups)).getPerformance();
+      // double hostPerformance =
+      // appInfoSuitableOptions.getCloudCharacteristics(elementName,
+      // YAMLmodulesOptimizerParser.getMeasuredPerformanceHost(elementName,
+      // groups)).getPerformance();
```
kiuby88: If it is not necessary, It could be better to avoid commented code.

perezp: @kiuby88 . Glups, I forgot to remove it. Thank you

planner/optimizer/optimizer-core/src/main/java/eu/seaclouds/platform/planner/optimizer/util/YAMLmatchmakerToOptimizerParser.java
rosogon: sl4fj adds a nice feature to avoid the concatenation of strings and the need to use isXXXEnabled() to improve the performance. Instead of using:
```log.debug("responsetime = " + responsetime + " ms");```

You can use:
```log.debug("responsetime = {} ms", responsetime);

The substitution is made internally only if the log is efective :)

planner/optimizer/optimizer-core/src/main/java/eu/seaclouds/platform/planner/optimizer/util/DefaultConstants.java
kiuby88: Sorry @perezp. It was my fault,  you are using loaders correctly.

kiuby88:
It is ok for me

planner/optimizer/optimizer-core/src/main/java/eu/seaclouds/platform/planner/optimizer/util/YAMLmatchmakerToOptimizerParser.java
perezp: @rosogon Thank you!

perezp:
@adriannieto  @kiuby88 @rosogon . 
Thank you for your comments. I think I've addressed all of them. 

I answered most of them after your comments. 
@adriannieto I've removed the class Reoptimizer because it contained no useful code at present. 
@rosogon I've followed your suggestion regarding the logging feature in some other places along the code... but I've not done it comprehensively yet. 

I've already squashed and pushed the changes

PR #217, opened 2016-01-24
Adding a set of DAMS which were made manually

kiuby88:
Some DAMS


perezp:
+1 
Thanks @kiuby88 .

PR #227, opened 2016-02-02
Fixed wrong ADP nodes prefixes in the monitoring-dam-generator.

MicheleGuerriero:
This PR fixes an issue in the monitor-dam-generator related to node templates in the ADP of type seaclouds.node.Compute or seaclouds.node.Platform that was previously defined to be seaclouds.Node.Compute, sea clouds.Node.Platform. The fix is applied also in the resources used by the planner for its tests.
@PaoloCifariello can you have a look at this?


PaoloCifariello:
Thanks @MicheleGuerriero, it is fine to me. 
+1

Issue #229, opened 2016-02-03
A node template type mismatch between ADP and AAM produces wrong DAM.  This is related to PR #234


adriannieto:
I found an issue regarding the database types, more precisely in the ADP .  As you can see in the AAM we define a web/db application made of web tier and a DB. How ever when the ADP is generated the node template type field `sc_req.db` is translated from:

```yaml
sc_req.db:
      derived_from: seaclouds.nodes.database.mysql.MySqlNode
```

into:

```yaml
  sc_req.db:
    derived_from: seaclouds.nodes.ControlledDynamicWebAppCluster
```

which is obviously wrong. Could @rosogon @perezp  @PaoloCifariello @kiuby88  take a look of this? It seems to me a mismatch between modules.

**AAM**:
```yaml
tosca_definitions_version: tosca_simple_yaml_1_0_0_wd03
description: Webchat
imports:
- tosca-normative-types:1.0.0.wd06-SNAPSHOT
topology_template:
  node_templates:
    web:
      type: sc_req.web
      properties:
        language: JAVA
        autoscale: false
        java_version:
          constraints:
          - greater_or_equal: '7'
          - less_or_equal: '7'
      requirements:
      - endpoint: db
    db:
      type: sc_req.db
      properties:
        autoscale: false
        mysql_version:
          constraints:
          - greater_or_equal: '5'
          - less_or_equal: '5'
node_types:
  sc_req.web:
    derived_from: seaclouds.nodes.webapp.jboss.JBoss7Server
    properties:
      java_support:
        constraints:
        - equal: true
      jboss_support:
        constraints:
        - equal: true
      java_version:
        constraints:
        - greater_or_equal: '7'
        - less_or_equal: '7'
  sc_req.db:
    derived_from: seaclouds.nodes.database.mysql.MySqlNode
    properties:
      mysql_support:
        constraints:
        - equal: true
      mysql_version:
        constraints:
        - greater_or_equal: '5'
        - less_or_equal: '5'
groups:
  operation_web:
    members:
    - web
    policies:
    - dependencies:
        operation_db: '29'
    - AppQoSRequirements:
        response_time:
          less_than: 150.0 ms
        availability:
          greater_than: 0.99
        cost:
          less_or_equal: 1000.0 euros_per_month
        workload:
          less_or_equal: 24000.0 req_per_min
  operation_db:
    members:
    - db
    policies:
    - dependencies: {}
```
**ADP**:
```yaml
[...]

PR #234, opened 2016-02-04
Solving node template type mismatch

perezp:
This PR solves the issue #229 . 

- The main contribution is to remove the line of code that changed the node types. Then, it also
- removes the tests that checked whether the node types had changed (this issue had not been accidental, but at some point we had thought that it was the correct solution)
- Adds some log messages
- Adds a file with a test AAM

@adriannieto Since you raised issue #229, could you check this PR?

### DIFF
--- a/planner/optimizer/optimizer-core/src/main/java/eu/seaclouds/platform/planner/optimizer/OptimizerInitialDeployment.java
+++ b/planner/optimizer/optimizer-core/src/main/java/eu/seaclouds/platform/planner/optimizer/OptimizerInitialDeployment.java
@@ -262,7 +262,10 @@ public class OptimizerInitialDeployment {
             log.debug("Adding upperWkl= " + upperWklBound + " lowerWkl=" + lowerWklBound + " poolsize=" + maxPoolSize);
             YAMLoptimizerParser.addScalingPolicyToModule(entry.getKey(), baseAppMap, lowerWklBound, upperWklBound, 1,
                   maxPoolSize);
-            YAMLoptimizerParser.changeModuleToScalableType(entry.getKey(), baseAppMap);
+            
+            //Here it is the point where the module type changed to be scalable (Dynamic Cluster, through changeModuleToScalableType method). 
+            //It looks like it is not required to change the type. TODO: Study this in more depth.
+            
          } else {
             log.debug("Module " + entry.getKey() + " was not scalable");
          }
--- a/planner/optimizer/optimizer-core/src/main/java/eu/seaclouds/platform/planner/optimizer/util/YAMLgroupsOptimizerParser.java
+++ b/planner/optimizer/optimizer-core/src/main/java/eu/seaclouds/platform/planner/optimizer/util/YAMLgroupsOptimizerParser.java
@@ -68,7 +68,9 @@ public class YAMLgroupsOptimizerParser {
    }
 
    public static Map<String, Object> getQoSinfoOfMemberName(String moduleName, Map<String, Object> groups) {
+      log.debug("Looking for QoS information benchmark of module {}. Finding its group", moduleName);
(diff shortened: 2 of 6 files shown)


planner/optimizer/optimizer-core/src/main/java/eu/seaclouds/platform/planner/optimizer/OptimizerInitialDeployment.java
```diff
@@ -262,7 +262,10 @@ private void addWorkloadAndPoolSizeBoundsForScalableModules(Solution solution,
             log.debug("Adding upperWkl= " + upperWklBound + " lowerWkl=" + lowerWklBound + " poolsize=" + maxPoolSize);
             YAMLoptimizerParser.addScalingPolicyToModule(entry.getKey(), baseAppMap, lowerWklBound, upperWklBound, 1,
                   maxPoolSize);
-            YAMLoptimizerParser.changeModuleToScalableType(entry.getKey(), baseAppMap);
```
adriannieto: What happens with the results if this is removed? No more scalable types?

adriannieto:
It looks good enough to merge but I'm curious about how your changes solved the issue.

planner/optimizer/optimizer-core/src/main/java/eu/seaclouds/platform/planner/optimizer/OptimizerInitialDeployment.java
perezp: If this method is removed, the types of modules remain unaltered from the declared in the AAM. 

"Scalable type" was for example the DynamicWebAppCluster. 
Some module (of any type, for example database.mysql.MySqlNode) are marked in the AAM as scalable, and the output of the optimizer can give the type of resources to use and the conditions when the module should scale-in/out. 
Once the previous conditions are computed, the type of the modules which got these scaling conditions was changed on purpose to a scalable type (this is, to the ControlledDynamicWebAppCluster). 

This behavior was not a bug or accidental, but it was done on purpose (it was not a bug, it was a feature). Actually, it had its own tests that checked that the type had been correctly changed. During the DAM learning process, in some moment we had thought that this was the suitable solution to implement. 

Since the change of the types declared in the AAM was done in method changeModuleToScalableType, just  removing the line of the invocation to this method solves the issue of the changes of types. 

Then, additionally, I removed the tests that checked that the types of modules had been correctly changed because now they are not changed. 

Does it answer your questions?

adriannieto:
The problem of #229 was not about changing the types, but changing to the right ones. The issue was that during the generation of the necessary files we were changing a MySQL  to a WebCluster, which was obviously wrong. 

I think that having the scalable option is a nice addition, so fixing the issue should focus on improve the non-scalable -> scalable mapping instead of removing it.

perezp:
@adriannieto I've understood your comment. Thank you. 
Now, I've been discussing and learning about the modification of types and the information on them. 

The mapping you have proposed  from non-scalable to scalable types is OK and will also entail adding the correct property values that "scalable" types require, because scalable types (like different kinds of clusters) have different properties from the MySQL of Jboss7server types. This information is out of hand of the information managed by the optimizer. So it will not be able to change it to the correct scalable type with the appropriate value for its properties. 

Therefore, I understand that this change of types from non-scalable to scalable should be done in other part of the code. I'm going to close the PR. Thus, types in the ADP will be still changed from an incorrect type (because they are non-scalable) to another incorrect type (scalable but yet incorrect).

When we know what is the suitable way to solve this issue I can reopen the PR to make the optimizer keep the same types as it has received as input, which is what this PR does.

perezp:
PR reopened for avoiding type change in the optimizer.

kiuby88:
I agree with this. When DamGenerator will work we will try to include the dynamic types management.
Thanks a lot @perezp

PR #235, opened 2016-02-04
Add default internet latency

perezp:
It was considered useful to have in the Optimizer a default value for the internet latency ( >0 ) in case that such value will not be configured


rosogon:
Is it possible to have real latency values between data centers?

perezp:
I don't know, but it would be wonderful to use real values. 

For what concerns this PR, this latency values are not "final", it's just a variable that can be set. 
At present, it is not being set; so it is better to have a default value of 200ms instead of a 0ms. 
It is not a final value anyway.

kiuby88:
This PR looks good. can we merged it?

perezp:
Thank you @kiuby88 . I'll merge it ASAP. I'm waiting until we have a clean master and we can execute travis before merging more PR.

PR #237, opened 2016-02-08
FEATURE(credentials): Add credentials forms and credentials storage according to provider.

rosogon:
This PR adds forms to enter the credentials for amazon, openshift and cloudfoundry. Another generic form (to simply type key/values separated by ';') is implemented in case the provider is not recognized. The credentials are stored in the brooklyn_location policy according the properties defined in section 7.2.2 of DAM model

It is important that the location property of the node_template related to the provider matches the variable ```form_by_provider``` (credential.js:23). I would need @kiuby88 and @jacopogiallo to check this.

@adriannieto : Could you check the PR? Thanks

### DIFF
--- a/dashboard/src/main/resources/webapp/static/lib/angular-topology-editor/credentials.js
+++ b/dashboard/src/main/resources/webapp/static/lib/angular-topology-editor/credentials.js
@@ -14,6 +14,18 @@
 
 var Credentials = (function() {
 
+    var activeform = undefined;
+    var amazonform = Object.create(Forms.Form);
+    var userpwdform = Object.create(Forms.Form);
+    var cloudfoundryform = Object.create(Forms.Form);
+    var genericform = Object.create(Forms.Form);
+
+    var form_by_provider = {
+        "aws-ec2": amazonform,
+        "openshift": userpwdform,
+        "CloudFoundry": cloudfoundryform,
+    };
+
     var canvas;
 
     /*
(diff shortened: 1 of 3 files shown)


dashboard/src/main/resources/webapp/static/lib/angular-topology-editor/credentials.js
```diff
@@ -14,6 +14,18 @@
...
+    var userpwdform = Object.create(Forms.Form);
+    var cloudfoundryform = Object.create(Forms.Form);
+    var genericform = Object.create(Forms.Form);
+
+    var form_by_provider = {
+        "aws-ec2": amazonform,
+        "openshift": userpwdform,
+        "CloudFoundry": cloudfoundryform,
+    };
+
```
jacopogiallo: It seems ok to me. @PaoloCifariello and @kiuby88 can provide more fruitful comments, as they are much more familiar than me about these "provider identifiers".

PaoloCifariello:
@rosogon "aws-ec2" and "CloudFoundry" are ok, but "openshift" is not present yet. I will add that value as location property of OpenShift offerings ASAP.

rosogon:
Ok, @PaoloCifariello . Let me know the final value of the openshift provider and I'll update the variable if needed.

adriannieto:
@rosogon It looks good for me, but I'd wait to @PaoloCifariello changes to merge it.

PaoloCifariello:
@adriannieto I already made the changes regarding "openshift" location, so I think you can merge it.

PR #241, opened 2016-02-11
FEATURE(aamwriter): Handles property application_requirements/frontend

rosogon:
This PR allows the specification of a frontend module in application_requirements/frontend element. Before this, the frontend module is calculated as the source node of the flow graph.

If the frontend (the node "visible" to the users, to which we want to apply the QoS requirements) is not the source of the node graph (assume a test module that connects to the frontend), the frontend must be specified.

Example:

```
{
    "nodes": [ ... ],
    "links": [ ... ],
    "application_requirements": {
        "response_time": "2000",
        "availability": 0.98,
        "cost": "200",
        "workload": "50",
        "frontend": "www"
    }
}
```

### DIFF
--- a/planner/aamwriter/src/main/java/eu/seaclouds/platform/planner/aamwriter/Translator.java
+++ b/planner/aamwriter/src/main/java/eu/seaclouds/platform/planner/aamwriter/Translator.java
@@ -239,9 +239,9 @@ public class Translator {
             sourceOperation.addDependencies(new Policy.Dependencies(dependencies));
             
             /*
-             * Application QosRequirements if this node is the source of the flow graph
+             * Application QosRequirements if this node is the frontend
              */
-            if (isSource(dgraph, dnode)) {
+            if (dnode.isFrontend()) {
                 DRequirements dreq = dgraph.getRequirements();
                 Policy.AppQoSRequirements requirements = new Policy.AppQoSRequirements(
                         dreq.getResponseTime(), dreq.getAvailability(), dreq.getCost(), dreq.getWorkload());
@@ -297,11 +297,6 @@ public class Translator {
         return artifactName;
     }
     
-    private boolean isSource(DGraph dgraph, DNode dnode) {
-        List<DLink> links = dgraph.getLinksTo(dnode);
-        return links.size() == 0;
-    }
-
     private String buildOperationName(DNode dnode) {
         return OPERATION_PREFIX + dnode.getName();
     }
--- a/planner/aamwriter/src/main/java/eu/seaclouds/platform/planner/aamwriter/modeldesigner/DGraph.java
+++ b/planner/aamwriter/src/main/java/eu/seaclouds/platform/planner/aamwriter/modeldesigner/DGraph.java
@@ -26,6 +26,17 @@ import org.json.simple.JSONObject;
 
 import eu.seaclouds.platform.planner.aamwriter.AamWriterException;
 
+/**
+ * Classes in aamwriter.modeldesigner package model the topology described in UI. 
+ * 
+ * DGraph class is the root of the topology model. It is in charge of reading a description
+ * of the topology in json and build the hierarchy of classes.
+ * 
+ * NOTE: The preferred way to set the frontend node of the application is specifying it in the 
+ * application_requirements element of the json topology. It not specified, the frontend
+ * node is calculated as the first node found without incoming links. This may lead to 
+ * unexpected behaviour. Due to optimizer constraint, only one frontend node is allowed.
+ */
 public class DGraph {
     public final static class Properties {
         public static final String NAME = "name";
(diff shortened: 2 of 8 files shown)


rosogon:
@adriannieto , @jacopogiallo : Could you have a look? Thanks

adriannieto:
@rosogon I don't see any change on the wizard javascript, isn't it necessary?

rosogon:
@adriannieto : this PR only covers the aamwriter. The changes in UI will be in a future PR.

planner/aamwriter/src/main/java/eu/seaclouds/platform/planner/aamwriter/modeldesigner/DRequirements.java
```diff
@@ -42,13 +44,22 @@ public DRequirements(JSONObject jnode, DGraph graph) {
...
+        double result = (value == null)? 0: Double.parseDouble(value.toString());
         
         return result;
     }
+    
+    private String parseString(JSONObject jnode, String key) {
+        Object value = jnode.get(key);
+        
+        String result = (value == null)? "" : value.toString();
+        return result;
```
jacopogiallo: Minor: Is it needed to declare result?

jacopogiallo:
@rosogon Could you please comment the code? What are DGraph, DNode, DRequirements, ... (and so on)?

planner/aamwriter/src/main/java/eu/seaclouds/platform/planner/aamwriter/modeldesigner/DGraph.java
```diff
@@ -122,13 +127,27 @@ public DNode getNode(String nodeName) {
...
+        for (DNode node : nodes) {
+            List<DLink> links = getLinksTo(node);
+            if (links.size() == 0) {
+                result = node;
+                break;
+            }
+        }
+        return result;
+    }
+
```
jacopogiallo: Just a comment: Here you are assuming that every application has exactly one node with 0 incoming links. Is this enforced by the GUI? I mean is the GUI forcing users to develop application that:
- cannot have all nodes with incoming dependecies (i.e. are all application directed acyclic graphs?), and
- cannot have two separate, independent stacks?

If not, you may encounter issues with this code.

planner/aamwriter/src/main/java/eu/seaclouds/platform/planner/aamwriter/modeldesigner/DRequirements.java
rosogon: No. You can define the method in one line, or not even using a method at all. But if you need to debug the method to see what value is taking each step, you will need to modify the code. I also prefer to have a single return point, with no conditions in it.

planner/aamwriter/src/main/java/eu/seaclouds/platform/planner/aamwriter/modeldesigner/DGraph.java
rosogon: @jacopogiallo : This is related with how the optimizer works. The optimizer assumes one single frontend (the one annotated with QoSAppRequirements). Chris is using such a graph, so I am adding the feature to select which node is the frontend from the UI (this PR is just the backend step). 

In summary, calculateSourceNode() is a fallback method, to avoid the aam generation fails. This worth a comment, indeed.

planner/aamwriter/src/main/java/eu/seaclouds/platform/planner/aamwriter/modeldesigner/DRequirements.java
jacopogiallo: +1

planner/aamwriter/src/main/java/eu/seaclouds/platform/planner/aamwriter/modeldesigner/DGraph.java
jacopogiallo: @rosogon I agree with you. Could you please add such a comment. :)

planner/aamwriter/src/main/java/eu/seaclouds/platform/planner/aamwriter/Translator.java
```diff
@@ -297,9 +297,8 @@ private String buildArtifactName(DNode dnode, boolean needsContainer) {
         return artifactName;
     }
     
-    private boolean isSource(DGraph dgraph, DNode dnode) {
-        List<DLink> links = dgraph.getLinksTo(dnode);
-        return links.size() == 0;
+    private boolean isFrontend(DGraph dgraph, DNode dnode) {
+        return dnode.equals(dgraph.getFrontendNode());
```
jacopogiallo: MINOR: Perhaps this could be a method of "DNode", something like "dnode.isFrontend(dgraph)".

rosogon:
@jacopogiallo : I hope I addressed your comments with the new commit. Thanks

jacopogiallo:
@rosogon +1! This PR is ok for me. :)

Issue #240, opened 2016-02-10
5 identical results in the "Optimize & Plan" step


rsucasas:
I get 5 identical results in the "Optimize & Plan" step.

AAM:
```
tosca_definitions_version: tosca_simple_yaml_1_0_0_wd03
description: Softcare
imports:
- tosca-normative-types:1.0.0.wd06-SNAPSHOT
topology_template:
  node_templates:
    forum-webapp-20151021:
      type: sc_req.forum-webapp-20151021
      properties:
        language: JAVA
        autoscale: false
node_types:
  sc_req.forum-webapp-20151021:
    derived_from: seaclouds.nodes.webapp.tomcat.TomcatServer
    properties:
      java_support:
        constraints:
        - equal: true
      tomcat_support:
        constraints:
        - equal: true
      java_version:
        constraints:
        - greater_or_equal: '7'
        - less_or_equal: '8'
      resource_type:
        constraints:
        - equal: platform
groups:
  operation_forum-webapp-20151021:
    members:
    - forum-webapp-20151021
    policies:
    - QoSInfo:
        execution_time: 100 ms
        benchmark_platform: Amazon_EC2_m1_medium_us_east_1
    - dependencies: {}
    - AppQoSRequirements:
        response_time:
          less_than: 200.0 ms
        availability:
          greater_than: 0.99
        cost:
          less_or_equal: 1000.0 euros_per_month
        workload:
          less_or_equal: 2400.0 req_per_min
```

These are the ADPs I get:
```
["tosca_definitions_version: tosca_simple_yaml_1_0_0_wd03
description: Softcare
imports:
- tosca-normative-types:1.0.0.wd06-SNAPSHOT
topology_template:
  node_templates:
    forum-webapp-20151021:
      type: sc_req.forum-webapp-20151021
      properties:
        language: JAVA
        autoscale: false
      requirements:
      - host: Cloud_Foundry
        instancesPOC: 1
    Cloud_Foundry:
      type: seaclouds.nodes.Platform.Cloud_Foundry
      properties:
        node_support: true
        go_support: true
        tomcat_version: 7
        private_hosting: true
        resource_type: platform
        tomcat_support: true
        auto_scaling: false
        public_hosting: false
        java_support: true
        horizontal_scaling: true
        ruby_support: true
        python_support: true
        scala_support: true
        vertical_scaling: true
        location: CloudFoundry
        go_version: 1.4
        php_support: true
        performance: 192
[...]

PR #242, opened 2016-02-11
Remove duplicate ADPs. Solving issue #240

perezp:
This PR avoids returning replicated ADPs in case that there exist just a very few cloud combinations. 

It is intended to solve issue #240 

It includes tests for the AAM that raised the issue; an AAM with only one module and just one cloud offer as possibility.

### DIFF
--- a/planner/optimizer/optimizer-core/src/main/java/eu/seaclouds/platform/planner/optimizer/OptimizerInitialDeployment.java
+++ b/planner/optimizer/optimizer-core/src/main/java/eu/seaclouds/platform/planner/optimizer/OptimizerInitialDeployment.java
@@ -108,9 +108,9 @@ public class OptimizerInitialDeployment {
          int numPlansToGenerate, double hyst) {
 
       log.debug("Optimization method started. Inputs received");
-      log.debug("AAM is: " + appModel);
-      log.debug("Suitable offers are: " + suitableCloudOffer);
-      log.debug("Informaton of Benchmark platform is: " + benchmarkPlatformsYaml);
+      log.trace("AAM is: {}", appModel);
+      log.trace("Suitable offers are: {} ", suitableCloudOffer);
+      log.trace("Informaton of Benchmark platform is: {}" , benchmarkPlatformsYaml);
 
       // Get app characteristics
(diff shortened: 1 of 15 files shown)


planner/optimizer/optimizer-core/src/main/java/eu/seaclouds/platform/planner/optimizer/OptimizerInitialDeployment.java
```diff
@@ -108,9 +108,9 @@ private String openBenchmarkOffersDefaultFile() {
          int numPlansToGenerate, double hyst) {
 
       log.debug("Optimization method started. Inputs received");
-      log.debug("AAM is: " + appModel);
-      log.debug("Suitable offers are: " + suitableCloudOffer);
-      log.debug("Informaton of Benchmark platform is: " + benchmarkPlatformsYaml);
+      log.debug("AAM is: {}", appModel);
```
andreaturli: - try to avoid multi-lines logs. 
- `log.info` is more [appropriate](http://stackoverflow.com/questions/6860186/log-info-vs-log-debug)

planner/optimizer/optimizer-core/src/main/java/eu/seaclouds/platform/planner/optimizer/heuristics/AbstractHeuristic.java
```diff
@@ -34,9 +34,9 @@
 
    static Logger log = LoggerFactory.getLogger(AbstractHeuristic.class);
 
-   private int MAX_ITER_NO_IMPROVE = 200; // 200;
-   private double MAX_TIMES_IMPROVE_REQUIREMENT = 20;
-   private static final int DEFAULT_MAX_NUM_INSTANCES = 10;
+   private int              MAX_ITER_NO_IMPROVE           = 200; // 200;
```
andreaturli: weird indentation

perezp:
Thank you @andreaturli . 
I've addressed both comments and updated the PR (although I'll take care of the indentation settings in the IDE, the issue about the indentation in attributes definition is prone to be repeated in the future).

andreaturli:
lgtm

PR #243, opened 2016-02-12
FIX(credentials): add support of new cloud type prefixes

rosogon:
This PR handles the prefix `tosca.nodes.(Compute|Platform)` as the new prefix for cloud types. This change in the planner is still not on master, so this PR still handles `seaclouds.nodes.(Compute|Platform)` as valid prefixes.

### DIFF
--- a/dashboard/src/main/resources/webapp/static/lib/angular-topology-editor/credentials.js
+++ b/dashboard/src/main/resources/webapp/static/lib/angular-topology-editor/credentials.js
@@ -253,14 +253,19 @@ var Credentials = (function() {
      * Returns a topology from a DAM.
      */
     function to_topology(rawdam) {
-        var PAAS_TYPE_PREFIX = "seaclouds.nodes.Platform";
-        var IAAS_TYPE_PREFIX = "seaclouds.nodes.Compute";
-        var INSTANCES = "instancesPOC";
+        var cloud_prefixes = [
+                "seaclouds.nodes.Platform",
+                "tosca.nodes.Platform",
+                "seaclouds.nodes.Compute",
+                "tosca.nodes.Compute"
+        ];
         var dam = jsyaml.safeLoad(rawdam);
 
         function node_type_is_provider(node_type_name) {
-            return node_type_name.startsWith(PAAS_TYPE_PREFIX) ||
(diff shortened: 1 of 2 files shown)


rosogon:
@kiuby88 , @adriannieto : Could you review this?

rosogon:
@kiuby88 , @adriannieto, @PaoloCifariello, @jacopogiallo  : Could you review this?

dashboard/src/main/resources/webapp/static/lib/angular-topology-editor/credentials.js
```diff
@@ -253,14 +253,19 @@ var Credentials = (function() {
...
     function to_topology(rawdam) {
-        var PAAS_TYPE_PREFIX = "seaclouds.nodes.Platform";
-        var IAAS_TYPE_PREFIX = "seaclouds.nodes.Compute";
-        var INSTANCES = "instancesPOC";
+        var cloud_prefixes = [
+                "seaclouds.nodes.Platform",
+                "tosca.nodes.Platform",
+                "seaclouds.nodes.Compute",
+                "tosca.nodes.Compute"
+        ];
```
jacopogiallo: seaclouds.nodes.XXXX are going to be removed in a future PR, right?

jacopogiallo:
@rosogon It seems ok for me. Have you tested that the generated AAM is the same that you were producing before (except for the possibility of specifying different types)?

dashboard/src/main/resources/webapp/static/lib/angular-topology-editor/credentials.js
rosogon: Yes. We can consider this PR as an intermediate step before that removal. After it, I can submit another PR to delete seaclouds.nodes.XXX from the array.

rosogon:
@jacopogiallo. It was not clear from the PR description. This PR is about detecting which nodes are clouds in the DAM, in the credentials screen. If a node is not identified as a cloud, it appears as a module, and the credentials for the cloud cannot be entered.

adriannieto:
LGTM the legacy features could be removed later when we have the final description of the DAM

rosogon:
Thanks, @jacopogiallo and @adriannieto . Merging

PR #245, opened 2016-02-16
Fix shell  DC's property

kiuby88:
This is the most important and longest commit


kiuby88:
@PaoloCifariello  take a look please

PaoloCifariello:
Thanks @kiuby88, it's ok to me! +1

PR #252, opened 2016-02-23
Feature/seaclouds initializer policy

adriannieto:
This policy is on charge to configure and clean up all the required SeaClouds Components (SLA Core, Tower4Clouds, InfluxDB observers and Grafana Dashboards)  when an application is installed/removed.

This policy is designed to be attached at root level. In order to see working examples please take a look to SeaCloudsInitializerPolicyTest.

It also remove unnecessary Java code for the dashboard.

DAM EXAMPLE:

```yaml
tosca_definitions_version: tosca_simple_yaml_1_0_0_wd03

imports:
  - tosca-normative-types:1.0.0.wd06-SNAPSHOT

template_name: brooklyn.a4c.simple.chatApplication
template_version: 1.0.0-SNAPSHOT

description: Brooklyn HelloWorld application. (ChatApplication)

node_types:
  org.apache.brooklyn.entity.webapp.tomcat.TomcatServer:
    derived_from: tosca.nodes.Root
    description: >
      A simple Tomcat server
    properties:
      wars.root:
        type: string
        required: false
      http.port:
        type: list
        required: false
        entry_schema:
          type: string
      java.sysprops:
        type: map
        required: false
        entry_schema:
          type: string
  org.apache.brooklyn.entity.database.mysql.MySqlNode:
    derived_from: tosca.nodes.Root
    description: >
      A MySQL server
    properties:
      "datastore.creation.script.url":
        type: string
        required: false
    capabilities:
      host:
        type: tosca.capabilities.Container

  seaclouds.nodes.Datacollector:
    derived_from: tosca.nodes.Root
    description: >
      A simple DC
    properties:
      installLatch:
        type: string
        required: false
      shell.env:
        type: map
        required: false
        entry_schema:
          type: string
    requirements:
    - host: tosca.nodes.Root
      type: tosca.relationships.HostedOn

  seaclouds.nodes.SlaInitializer:
    derived_from: tosca.nodes.Root
    description: >
      A node to initialize the SLA Core for a particular app
    properties:
      installLatch:
        type: string
        required: true
      shell.env:
        type: map
        required: true
        entry_schema:
          type: string
    requirements:
    - host: tosca.nodes.Root
      type: tosca.relationships.HostedOn

  seaclouds.nodes.Tower4CloudsInitializer:
    derived_from: tosca.nodes.Root
    description: >
      A node to initialize the Tower4Clouds platform for a particular app
    properties:
      installLatch:
        type: string
        required: true
      shell.env:
        type: map
        required: true
[...]

PR #254, opened 2016-02-24
FIX(replacing offerings type during DAM generation)

PaoloCifariello:
This PR adds the code to:

- replace offerings type (seaclouds.nodes.Compute.xxx and seaclouds.nodes.Platform.xxx) with tosca.nodes.Compute or tosca.nodes.Platform
- remove the requirement {endpoint:"", type:""} that is not supported yet by the deployer

### DIFF
--- a/planner/core/src/main/java/eu/seaclouds/platform/planner/core/DamGenerator.java
+++ b/planner/core/src/main/java/eu/seaclouds/platform/planner/core/DamGenerator.java
@@ -13,7 +13,6 @@ import static com.google.common.base.Preconditions.*;
 import java.net.MalformedURLException;
 import java.net.URL;
 import java.util.*;
-import java.util.regex.Pattern;
 
 
 /**
@@ -46,6 +45,7 @@ public class DamGenerator {
     public static final String REGION = "region";
     public static final String HARDWARE_ID = "hardwareId";
     public static final String TYPE = "type";
+    public static final String ENDPOINT = "endpoint";
     public static final String CLOUD_FOUNDRY = "CloudFoundry";
     public static final String POLICIES = "policies";
     public static final String GROUPS = "groups";
@@ -85,6 +85,7 @@ public class DamGenerator {
         adpYaml = DamGenerator.manageTemplateMetada(adpYaml);
         adpYaml = DamGenerator.translateAPD(adpYaml);
         adpYaml = DamGenerator.addMonitorInfo(yml.dump(adpYaml), monitorGenURL, monitorGenPort, influxdbURL, influxdbPort);
+        adpYaml = DamGenerator.translateOfferingsType(adpYaml);
 
         String slaInfoResponse = new HttpHelper(slaGenURL).postInBody(SLA_GEN_OP, yml.dump(adpYaml));
         checkNotNull(slaInfoResponse, "Error getting SLA info");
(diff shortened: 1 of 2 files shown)


planner/core/src/main/java/eu/seaclouds/platform/planner/core/DamGenerator.java
```diff
@@ -85,6 +85,7 @@ public static String generateDam(String adp, String monitorGenURL, String monito
         adpYaml = DamGenerator.manageTemplateMetada(adpYaml);
```
andreaturli: I think this would be much easier if we follow the same pattern used by the [matchmaker](https://github.com/SeaCloudsEU/SeaCloudsPlatform/blob/master/planner/matchmaker/src/main/java/eu/seaclouds/planner/matchmaker/Matchmaker.java#L172) - wdyt?

andreaturli:
thanks @PaoloCifariello I'm not sure this goes in the same direction as [252](https://github.com/SeaCloudsEU/SeaCloudsPlatform/pull/252) 

if we merge 252, I think the DAMGenerator should be modified to append that additional policy to the final `DAM` and possibly we'd need to change again the DAMGenerator, wdyt?

planner/core/src/main/java/eu/seaclouds/platform/planner/core/DamGenerator.java
```diff
@@ -46,6 +45,7 @@
     public static final String REGION = "region";
     public static final String HARDWARE_ID = "hardwareId";
     public static final String TYPE = "type";
+    public static final String ENDPOINT = "endpoint";
     public static final String CLOUD_FOUNDRY = "CloudFoundry";
     public static final String POLICIES = "policies";
     public static final String GROUPS = "groups";
```
andreaturli: can you please add a javadoc to the generateDAM method?
this will help particulalry to know what is inside the String returned, thanks

planner/core/src/main/java/eu/seaclouds/platform/planner/core/DamGenerator.java
```diff
@@ -272,6 +283,27 @@ public static String generateDam(String adp, String monitorGenURL, String monito
...
+        Map<String, Object> topologyTemplate = (Map<String, Object>) adpYaml.get(TOPOLOGY_TEMPLATE);
+        Map<String, Object> nodeTemplates = (Map<String, Object>) topologyTemplate.get(NODE_TEMPLATES);
+
+        for(String moduleName:nodeTemplates.keySet()) {
+            Map<String, Object> module = (Map<String, Object>) nodeTemplates.get(moduleName);
+
+            String moduleType = (String) module.get("type");
+
+            // replaces seaclouds types (unknown by the deployer) with tosca types for the offerings
+            if (moduleType.startsWith("seaclouds.nodes.Platform.")) {
```
andreaturli: this looks hacky, better to understand which step is adding `type` with those prefix `seaclouds.nodes.Platform.` and fix that step properly.

planner/core/src/main/java/eu/seaclouds/platform/planner/core/DamGenerator.java
PaoloCifariello: Of course I can @andreaturli, thanks for your suggest.

PaoloCifariello:
Thanks @andreaturli. Actually this is just the "ongoing" version deployed on Amazon.

During this week I will work to modify the DAM generator in order to be aligned with #252 and I will PR it.

planner/core/src/main/java/eu/seaclouds/platform/planner/core/DamGenerator.java
PaoloCifariello: Yes, you are right. This type come out from the offerings fetched by the Discoverer. I am trying to understand if it is feasible to change that type in the very first step. In that case this part of the code will be no more needed.

andreaturli:
ok thanks @PaoloCifariello, better close this PR then?

rosogon:
@andreaturli , @PaoloCifariello : I would agree in changing seaclouds.nodes.(Platform|Compute) for tosca.nodes.\1, but tosca.nodes.Platform is not standard.

@PaoloCifariello : if you change the discoverer to use tosca.nodes instead of seaclouds.nodes, remember that the AAMWriter have to be changed, too.

andreaturli:
thanks @rosogon for the heads-up.

PR #258, opened 2016-02-26
Improving isRunning of PaaS entities

kiuby88:
PaaS entities will be running when application will be deployed on the Platform and the main url will be available


perezp:
It improves the current state of the code. Looks good to me.

PR #257, opened 2016-02-26
Avoiding NodeTypes in Sample DAMs

kiuby88:
DAMs do not require NodeTypes definitions. [SeaClouds NodeTypes](https://github.com/SeaCloudsEU/seaclouds-types) are pre-loaded in the SeaClouds Deployer.

PR #259, opened 2016-03-01
Feature(Planner): Removed node_types from DAM

PaoloCifariello:
This PR adds the code to remove node_types inside the DAM and to add the import containing such node_types.

@kiuby88 can you take a look?

### DIFF
--- a/planner/core/src/main/java/eu/seaclouds/platform/planner/core/DamGenerator.java
+++ b/planner/core/src/main/java/eu/seaclouds/platform/planner/core/DamGenerator.java
@@ -13,7 +13,6 @@ import static com.google.common.base.Preconditions.*;
 import java.net.MalformedURLException;
 import java.net.URL;
 import java.util.*;
-import java.util.regex.Pattern;
 
 
 /**
@@ -75,10 +74,22 @@ public class DamGenerator {
     public static final String SEACLOUDS_APPLICATION_INFORMATION_POLICY_TYPE = "seaclouds.policies.app.information";
     public static final String SEACLOUDS_APPLICATION_POLICY_NAME = "seaclouds.app.information";
     public static final String SEACLOUDS_NODE_PREFIX = "seaclouds.nodes";
+    private static final String SEACLOUDS_NORMATIVE_TYPES = "seaclouds-types";
+    private static final String SEACLOUDS_NORMATIVE_TYPES_VERSION = "0.8.0-SNAPSHOT";
 
     private static DeployerTypesResolver deployerTypesResolver;
     static Map<String, MonitoringInfo> monitoringInfoByApplication=new HashMap<>();
 
+    /**
+     *
+     * @param adp YAML string containing the ADP
+     * @param monitorGenURL URL of the Monitoring service
+     * @param monitorGenPort Port of the Monitoring service
+     * @param slaGenURL URL of the SLA service
+     * @param influxdbURL URL of InfluxDB
+     * @param influxdbPort Port of InfluxDB
+     * @return YAML string containing the generated DAM
+     */
     public static String generateDam(String adp, String monitorGenURL, String monitorGenPort, String slaGenURL, String influxdbURL, String influxdbPort){
         Yaml yml = new Yaml();
         Map<String, Object> adpYaml = (Map<String, Object>) yml.load(adp);
@@ -91,6 +102,7 @@ public class DamGenerator {
         adpYaml = DamGenerator.addApplicationInfo(adpYaml, slaInfoResponse, SLA_INFO_GROUPNAME);
 
         Map groups = (Map) adpYaml.remove(GROUPS);
+        adpYaml.remove(NODE_TYPES);
 
         addPoliciesTypeIfNotPresent(groups);
 
@@ -115,6 +127,8 @@ public class DamGenerator {
                     imports.remove(importedNormativeTypes);
                     imports.add(TOSCA_NORMATIVE_TYPES+":"+TOSCA_NORMATIVE_TYPES_VERSION);
                 }
+
+                imports.add(SEACLOUDS_NORMATIVE_TYPES + ":" + SEACLOUDS_NORMATIVE_TYPES_VERSION);
             }
         }
 
@@ -263,12 +277,9 @@ public class DamGenerator {
 
             policyGroup.put(POLICIES, policy);
 
-            HashMap<String, Object> finalGroup = new HashMap<>();
-
             ADPgroups.put(ADD_BROOKLYN_LOCATION + group ,policyGroup);
         }
 
-        String finalDam = yml.dump(adpYaml);
         return adpYaml;
     }
 
--- a/planner/core/src/main/java/eu/seaclouds/platform/planner/core/DamGenerator.java
+++ b/planner/core/src/main/java/eu/seaclouds/platform/planner/core/DamGenerator.java
@@ -74,8 +74,8 @@ public class DamGenerator {
     public static final String SEACLOUDS_APPLICATION_INFORMATION_POLICY_TYPE = "seaclouds.policies.app.information";
     public static final String SEACLOUDS_APPLICATION_POLICY_NAME = "seaclouds.app.information";
     public static final String SEACLOUDS_NODE_PREFIX = "seaclouds.nodes";
-    private static final String SEACLOUDS_NORMATIVE_TYPES = "seaclouds-types";
-    private static final String SEACLOUDS_NORMATIVE_TYPES_VERSION = "0.8.0-SNAPSHOT";
+    public static final String SEACLOUDS_NODE_TYPES = "seaclouds-types";
+    public static final String SEACLOUDS_NODE_TYPES_VERSION = "0.8.0-SNAPSHOT";
 
     private static DeployerTypesResolver deployerTypesResolver;
     static Map<String, MonitoringInfo> monitoringInfoByApplication=new HashMap<>();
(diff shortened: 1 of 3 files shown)


planner/core/src/main/java/eu/seaclouds/platform/planner/core/DamGenerator.java
```diff
@@ -75,10 +74,22 @@
     public static final String SEACLOUDS_APPLICATION_INFORMATION_POLICY_TYPE = "seaclouds.policies.app.information";
     public static final String SEACLOUDS_APPLICATION_POLICY_NAME = "seaclouds.app.information";
     public static final String SEACLOUDS_NODE_PREFIX = "seaclouds.nodes";
+    private static final String SEACLOUDS_NORMATIVE_TYPES = "seaclouds-types";
```
kiuby88: Probably it could be better something like `SEACLOUDS_NODE_TYPES`

kiuby88:
Thanks @PaoloCifariello, this PR completes the  initial step of #257 
It looks good but a lot of test should be necessary.

PaoloCifariello:
Thanks @kiuby88, I will add some tests ASAP.

planner/core/src/main/java/eu/seaclouds/platform/planner/core/DamGenerator.java
```diff
@@ -91,6 +102,7 @@ public static String generateDam(String adp, String monitorGenURL, String monito
         adpYaml = DamGenerator.addApplicationInfo(adpYaml, slaInfoResponse, SLA_INFO_GROUPNAME);
 
         Map groups = (Map) adpYaml.remove(GROUPS);
+        adpYaml.remove(NODE_TYPES);
```
andreaturli: why the ADP contains the NodeTypes?

planner/core/src/main/java/eu/seaclouds/platform/planner/core/DamGenerator.java
```diff
@@ -75,10 +74,22 @@
...
     public static final String SEACLOUDS_NODE_PREFIX = "seaclouds.nodes";
+    public static final String SEACLOUDS_NODE_TYPES = "seaclouds-types";
+    public static final String SEACLOUDS_NODE_TYPES_VERSION = "0.8.0-SNAPSHOT";
 
     private static DeployerTypesResolver deployerTypesResolver;
     static Map<String, MonitoringInfo> monitoringInfoByApplication=new HashMap<>();
 
+    /**
+     *
+     * @param adp YAML string containing the ADP
```
andreaturli: why a String? Can't you use the same approach used at https://github.com/SeaCloudsEU/SeaCloudsPlatform/blob/master/planner/matchmaker/src/main/java/eu/seaclouds/planner/matchmaker/Matchmaker.java#L172 ?

ADP is simply a TOSCA `ArchiveRoot`. wdyt?

planner/core/src/main/java/eu/seaclouds/platform/planner/core/DamGenerator.java
```diff
@@ -75,10 +74,22 @@
...
+     * @param monitorGenURL URL of the Monitoring service
+     * @param monitorGenPort Port of the Monitoring service
+     * @param slaGenURL URL of the SLA service
+     * @param influxdbURL URL of InfluxDB
+     * @param influxdbPort Port of InfluxDB
+     * @return YAML string containing the generated DAM
```
andreaturli: why another String? Isn't the `DAM` yet another `ArchiveRoot`?

planner/core/src/main/java/eu/seaclouds/platform/planner/core/DamGenerator.java
PaoloCifariello: It is not the ADP, but it is the partial dam. The name is misleading. I am going to change it.

andreaturli: whatever that is, why `NodeTypes` are there, if you then want to remove it?

PaoloCifariello: they come out from the AAM and are needed for adding the monitoring rules to the DAM. After that `NodeTypes` are no more needed because @kiuby88 included it in the Deployer.

kiuby88: Good question, @andreaturli.
I am not in charge of DamGenerator, but I will try to explain.
ADPs contain NodeTypes of SeaClouds which are used during AAM generation and ADP management. However, this type will be used by the Deployer. Current DamGenerator builds the DAM over the ADP object so the ADP's NodeTypes has to be removed. 
Of course, the best option it is refactor the current DamGenerator, and I think it must be done.

planner/core/src/main/java/eu/seaclouds/platform/planner/core/DamGenerator.java
kiuby88: I am not sure if current ADP can be managed as a [ArchiveRoot](https://github.com/alien4cloud/alien4cloud/blob/d802eef4cc3cd7416155d4716ecb79e1de861dd8/alien4cloud-core/src/main/java/alien4cloud/tosca/model/ArchiveRoot.java) but I think we have to try

planner/core/src/main/java/eu/seaclouds/platform/planner/core/DamGenerator.java
kiuby88: I am sure it is possible, because DAM will be processed by brooklyn-tosca which will use `ArchiveRoot`.

kiuby88: I think a new PR should be open in order to refactor DamGenerator

planner/core/src/main/java/eu/seaclouds/platform/planner/core/DamGenerator.java
andreaturli: @PaoloCifariello thanks, are you sure NodeType are required to generate the monitoring rules?

PaoloCifariello: I talked with Michele about that and he said that information is needed.
@MicheleGuerriero can you confirm?

MicheleGuerriero: Yes the monitoring-dam-generator looks at the type basically to infer things like if a module is a database pr not.

andreaturli: thanks @MicheleGuerriero. I assume you are aware of #252 and the long open discussion on the DAM and then the DamGenerator, yes?
I cannot see `monitor-dam-generator` used in #252. Please coordinate among yourselves (@MicheleGuerriero, @PaoloCifariello and @kiuby88 ) and when you have a clear understanding of the situation re-open those conflicting PRs. Thanks

kiuby88: Hi.

#257 and #259 are aligned. Both of them avoid NodeTypes, of course they are related between them. They allow to build and managed DAMs without NodeTypes definitions, and these DAMs can be deployed and it should work and they can work without policies so I think they are independent of #252. 

#252 adds policy and some Dashboard modifications, but we need to clarify the policy.  We need to clarify some policy points before adding to DAMs. #252 does not use `MonitoringDamGenerator` because the policy requires the generated MonitoringRules. When, we connected the `DamGenerator` and the Policy we will used `MonitoringDamGenerator`.

Of course, we could close #257 and #259 and move their changes to a new PR. Then, we will continue with policy.
WDYT?

MicheleGuerriero: @andreaturli yes I'm aware of #252 and of the many changes that is carries at this point. I think the MonitorDamGenerator is not affected by #252 since this PR just changes the way information produced by the monitorDamGenerator are used, but the information itself is unchanged. What I'm missing is how the removal of NodeTypes from the current ADP (actually, at least so far, the node types have always been there) is connected with #252. Can you clarify me this point? I'm asking because on the other hand the removal of node types from the ADP would instead affect the MonitorDamGenerator.

andreaturli: Don't think I can. I'm simply trying to look after the repo and asking questions to to have a discussion to understand a bit more the context.

kiuby88: @MicheleGuerriero, I think #252 does not affect `MonitorDamGenerator`.
Please, take a look [here](https://github.com/kiuby88/SeaCloudsPlatform/commit/504deab1a648be90e58672a2e4c1e60514875f06). I have added to DamGenerator a fist current policy integration. I hope it could be useful to clarify some points.

This code it will not push to SeaClouds (at least we need a minor refactoring before) and I am solving Andrea comments about the policy. But it is only a first example integration.

kiuby88: I think @MicheleGuerriero is using the ADP NodeTypes [here](https://github.com/kiuby88/SeaCloudsPlatform/blob/feature/seaclouds-policy-generation/monitor/monitoring-dam-generator-core/src/main/java/eu/seaclouds/monitor/monitoringdamgenerator/MonitoringDamGenerator.java#L103-L104), so we should not delete NodeTypes from ADP. Of course, we should avoid NodeTypes in final DAM, but ADP needs its own  NodeTypes.

andreaturli: @kiuby88, what do I think? :)
I think you guys should close #252 #257 and #259 and open a new PR that changes consistently the DAMGenerator to do whatever you want to achieve. My two cents

kiuby88:
Paolo, close this PR please

kiuby88:
Thanks @PaoloCifariello

PR #260, opened 2016-03-01
Adds the env var where an endpoint url is stored.

rosogon:
The link form in the designer asks for the env var.
The env var is introduced in the AAM at aamwriter time.

This also removes a previous feature where credentials were stored in a
file.

For example, the following topology:

```
    [...]
    "links": [
        {
            "source": "www",
            "target": "webservices",
            "properties": {
                "calls": "2",
                "operation_type": "seaclouds.relation.connection.endpoint.host",
                "env_var": "connurl"
            }
        },
    [...]
```

generates the following requirement in the node template:

```
    www:
      type: sc_req.www
      properties:
        language: JAVA
        location: DYNAMIC
        location_option: FOLLOW_SUN
        autoscale: true
      artifacts:
      - wars.root: http://www.example.com/downloads/www.war
        type: tosca.artifacts.File
      requirements:
      - endpoint: webservices
        type: seaclouds.relation.connection.endpoint.host
        properties:
          prop.name: connurl
```

@kiuby88 : Could you review?

### DIFF
--- a/dashboard/src/main/resources/webapp/static/lib/angular-topology-editor/editor-view.html
+++ b/dashboard/src/main/resources/webapp/static/lib/angular-topology-editor/editor-view.html
@@ -240,9 +240,9 @@
                                 <select class="form-control" id="operation-type"> </select>
                             </div>
                             <div class="form-group">
-                                <label for="operation-credentials-file">File where to store credentials</label>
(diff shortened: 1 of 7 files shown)


kiuby88:
Looks good

PR #262, opened 2016-03-04
Fix/dam generator refactoring

kiuby88:
A first refactoring of `DamGenerator`.
I have deleted static methods and I have added some little test.

This PR maintains the integrity of the current SeaClouds version. It only contains a little refactoring, while Dashboard is in charge of orchestrate the rest of components still. Then, sample DAMs have been updated according to this context. Moreover, this PRs contains #259 and #257 (which should already be closed).
@PaoloCifariello, could you take a look please?

Note:
I think that the best option to refactor `DamGenerator` is to use an object model, for example alien4cloud, but it is the first refactor version and...


PaoloCifariello:
Thanks @kiuby88 for this first refactor.
It is ok to me, good job. +1

rosogon:
@kiuby88 : Squash, please

kiuby88:
@rosogon done

PR #264, opened 2016-03-09
Test(Planner): Added test for DAM generator

PaoloCifariello:
I have added few tests for DAM generator.
Three DAMs are produced and compared with precomputed ones.

@kiuby88 can you please take a look?


planner/core/src/test/java/eu/seaclouds/platform/planner/core/DamGeneratorTest.java
```diff
@@ -124,5 +125,112 @@ public void testGroupsAsTopologyChild() throws Exception {
...
+
+        when(fakeAgreementManager.generateAgreeemntId(((Map<String, Object>) anyObject())))
+                .thenReturn(FAKE_AGREEMENT_ID);
+
+        DamGenerator damGenerator = new DamGenerator(MONITOR_URL, MONITOR_PORT, SLA_ENDPOINT, INFLUXDB_URL, INFLUXDB_PORT);
+        damGenerator.setAgreementManager(fakeAgreementManager);
+        dam = damGenerator.generateDam(adp);
+        template = (Map<String, Object>) yamlParser.load(dam);
+
+        String expectedDAMString = new Scanner(new File(Resources.getResource("nuro_dam.yml").toURI())).useDelimiter("\\Z").next();
```
kiuby88: When this dile is generate for this test?

planner/core/src/test/java/eu/seaclouds/platform/planner/core/DamGeneratorTest.java
```diff
@@ -124,5 +125,112 @@ public void testGroupsAsTopologyChild() throws Exception {
...
+        Map<String, Object> expectedDAM = (Map<String, Object>) yamlParser.load(expectedDAMString);
```
kiuby88: Why we have two different objects with the same information, `template` and `expectedDAM`?
Please, use `Dam`, take a look [here](http://stackoverflow.com/questions/15526107/acronyms-in-camelcase)

planner/core/src/test/java/eu/seaclouds/platform/planner/core/DamGeneratorTest.java
```diff
@@ -124,5 +125,112 @@ public void testGroupsAsTopologyChild() throws Exception {
...
+
+        assertNotNull(template);
+
+        Map<String, Object> generatedTopologyTemplate = (Map<String, Object>) template.get(DamGenerator.TOPOLOGY_TEMPLATE);
+        Map<String, Object> expectedTopologyTemplate = (Map<String, Object>) expectedDAM.get(DamGenerator.TOPOLOGY_TEMPLATE);
```
kiuby88: I think it is duplicated and it should be remove and using `template`

planner/core/src/test/java/eu/seaclouds/platform/planner/core/DamGeneratorTest.java
```diff
@@ -124,5 +125,112 @@ public void testGroupsAsTopologyChild() throws Exception {
...
+
+        assertEquals(generatedTopologyTemplate.get("www"), expectedTopologyTemplate.get("www"));
```
kiuby88: Probably it is better to compare the `template` elements with the values and to avoid using the expected DAM (`expectedTopology`). It is more clear.

planner/core/src/test/java/eu/seaclouds/platform/planner/core/DamGeneratorTest.java
```diff
@@ -124,5 +125,112 @@ public void testGroupsAsTopologyChild() throws Exception {
...
+        assertEquals(generatedTopologyTemplate.get("db"), expectedTopologyTemplate.get("db"));
+        assertEquals(generatedTopologyTemplate.get("Amazon_EC2_m1_small_eu_central_1"), expectedTopologyTemplate.get("Amazon_EC2_m1_small_eu_central_1"));
+        assertEquals(generatedTopologyTemplate.get("Amazon_EC2_m4_10xlarge_eu_west_1"), expectedTopologyTemplate.get("Amazon_EC2_m4_10xlarge_eu_west_1"));
+
+        Map<String, Object> generatedGroups = (Map<String, Object>) generatedTopologyTemplate.get(DamGenerator.GROUPS);
+        Map<String, Object> expectedGroups = (Map<String, Object>) expectedTopologyTemplate.get(DamGenerator.GROUPS);
+
+        assertEquals(generatedGroups.get("operation_www"), expectedGroups.get("operation_www"));
```
kiuby88: Each property should be tested. I think `tosca.nodes.Compute` NodeTypes properties could avoided in the test but the rest of the NodeTemplates should be reviewed.
Policies and their properties should also be tested.

planner/core/src/test/java/eu/seaclouds/platform/planner/core/DamGeneratorTest.java
```diff
@@ -124,5 +125,112 @@ public void testGroupsAsTopologyChild() throws Exception {
...
+        assertEquals(generatedGroups.get("operation_db"), expectedGroups.get("operation_db"));
+        assertEquals(generatedGroups.get("add_brooklyn_location_Amazon_EC2_m1_small_eu_central_1"), expectedGroups.get("add_brooklyn_location_Amazon_EC2_m1_small_eu_central_1"));
+        assertEquals(generatedGroups.get("add_brooklyn_location_Amazon_EC2_m4_10xlarge_eu_west_1"), expectedGroups.get("add_brooklyn_location_Amazon_EC2_m4_10xlarge_eu_west_1"));
+    }
+
+    @Test
+    @SuppressWarnings("unchecked")
+    public void testAtosDAM() throws Exception {
```
kiuby88: Follow the previous suggestions

planner/core/src/test/java/eu/seaclouds/platform/planner/core/DamGeneratorTest.java
```diff
@@ -124,5 +125,112 @@ public void testGroupsAsTopologyChild() throws Exception {
...
+        assertEquals(generatedGroups.get("operation_webservices"), expectedGroups.get("operation_webservices"));
+        assertEquals(generatedGroups.get("operation_db1"), expectedGroups.get("operation_db1"));
+        assertEquals(generatedGroups.get("add_brooklyn_location_Amazon_EC2_c1_xlarge_eu_central_1"), expectedGroups.get("add_brooklyn_location_Amazon_EC2_c1_xlarge_eu_central_1"));
+        assertEquals(generatedGroups.get("add_brooklyn_location_Amazon_EC2_t2_micro_us_east_1"), expectedGroups.get("add_brooklyn_location_Amazon_EC2_t2_micro_us_east_1"));
+        assertEquals(generatedGroups.get("add_brooklyn_location_Amazon_EC2_m4_large_eu_west_1"), expectedGroups.get("add_brooklyn_location_Amazon_EC2_m4_large_eu_west_1"));
+    }
+
+    @Test
+    @SuppressWarnings("unchecked")
+    public void testWebChatDAM() throws Exception {
```
kiuby88: Follow the previous suggestions

kiuby88:
Thanks a lot @PaoloCifariello, good start point.
I have added some comments.

Please, organize the useful ADPs in a folder and delete the unused files.

kiuby88:
Thanks for updating @PaoloCifariello 
But I think it is necessary to test the properties. For example, 
`assertEquals(nodeTemplate.getProperty("wars.root"), generatedNodeTemplate.getPtoperty("wars.root"))``

All policies should be tested too.

PaoloCifariello:
Thanks for you suggestion @kiuby88, but I already test all the properties. 
For example in https://github.com/SeaCloudsEU/SeaCloudsPlatform/pull/264/files#diff-b607e449996535837a5e3e8521d6646fR152  I assert that all the properties of the module named 'www' in the generated DAM are equal to the properties of the module named 'www' in the precomputed DAM (nuro_dam.yml).

kiuby88:
Hi @PaoloCifariello , probably I should have been more concrete. I tried to say that generated DAM should to be tested completely. For example, I think the following lines are not tested:
* https://github.com/PaoloCifariello/SeaCloudsPlatform/blob/test/dam-generator-test/planner/core/src/test/resources/atos/atos_dam.yml#L189-L197
* https://github.com/PaoloCifariello/SeaCloudsPlatform/blob/test/dam-generator-test/planner/core/src/test/resources/atos/atos_dam.yml#L31-L147

Those are some examples of important lines that should be checked (I have omitted some lines which are not important), could you take a look?
Thx

PaoloCifariello:
Thanks for your suggest @kiuby88, I agree that I need to test all the node templates, but I don't think we need to test the offerings node templates (https://github.com/PaoloCifariello/SeaCloudsPlatform/blob/test/dam-generator-test/planner/core/src/test/resources/atos/atos_dam.yml#L31-L45) because they are left untouched by the DAM generator.

The groups you provided (https://github.com/PaoloCifariello/SeaCloudsPlatform/blob/test/dam-generator-test/planner/core/src/test/resources/atos/atos_dam.yml#L189-L197) contain only an ID that is different each time you call `generateDam`. What kind of test can we make about it?

I am adding tests for all node templates (except the offerings) as you suggested, but I don't think we need much more, WDYT?

kiuby88:
Hi @PaoloCifariello 
Probably [monitoring info id](https://github.com/PaoloCifariello/SeaCloudsPlatform/blob/test/dam-generator-test/planner/core/src/test/resources/atos/atos_dam.yml#L191) is different but [sla_gen_info id](https://github.com/PaoloCifariello/SeaCloudsPlatform/blob/test/dam-generator-test/planner/core/src/test/resources/atos/atos_dam.yml#L196) is mocked so you could try to test it. In any case, as you have mentioned this pair of policies are not very important (although I'd like to check them).

About offering testing, it is mandatory. `DamGenerator` generates a DAM and we have to sure that all elements are generated how we expected. So please, include offerings

PaoloCifariello:
Thanks for the clarification @kiuby88, I have added a check also for the offerings.

kiuby88:
@perezp could you merge this PR?

kiuby88:
Thx

PR #261, opened 2016-03-03
Adds specification of frontend node in the designer.

rosogon:
A frontend node is specified in the topology as a boolean property in
the properties of a node. Visually, it appears with black rounded line.
The rest of nodes appears with a grey rounded line. Only one frontend
node can be specified.

This way of specifying a frontend has less priority than the previous
one when generating the AAM (setting the frontend node in the
app_requirements element), but it is easier to implement from the designer.

Form code for the topology editor has been moved from editor.js to
editor-forms.js, as the former file was too complex.

PR #265, opened 2016-03-10
FEATURE: optimizer accepts topologies where not all modules are reached from the frontend

perezp:
This PR complements #261 , on the optimizer side. 

The last topologies seen included modules that were not reached by the frontend (admin-interface, simulator). This was not supported by the optimizer, nor the GUI. 
This PR makes the optimizer support this kind of topology, #261 made the GUI support the explicit assignment of frontend. 

An example with a TOSCA topology with a disconnected module is added in this PR, and tests are included using such TOSCA topology

### DIFF
--- a/planner/optimizer/optimizer-core/src/main/java/eu/seaclouds/platform/planner/optimizer/heuristics/BlindSearch.java
+++ b/planner/optimizer/optimizer-core/src/main/java/eu/seaclouds/platform/planner/optimizer/heuristics/BlindSearch.java
@@ -17,6 +17,8 @@
 
 package eu.seaclouds.platform.planner.optimizer.heuristics;
 
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 import eu.seaclouds.platform.planner.optimizer.Solution;
 import eu.seaclouds.platform.planner.optimizer.SuitableOptions;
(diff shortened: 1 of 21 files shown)


planner/optimizer/optimizer-core/src/main/java/eu/seaclouds/platform/planner/optimizer/heuristics/BlindSearch.java
```diff
@@ -33,13 +37,12 @@ public BlindSearch() {
...
-   
-   
+
    @Override
-   public Solution[] computeOptimizationProblemForAllDifferentSolutions(SuitableOptions cloudOffers, QualityInformation requirements,
-         Topology topology, int numPlansToGenerate) {
-      return super.filterUniqueSolutions(computeOptimizationProblem(cloudOffers,requirements,topology,numPlansToGenerate));
+   public Solution[] computeOptimizationProblemForAllDifferentSolutions(SuitableOptions cloudOffers,
+         QualityInformation requirements, Topology topology, int numPlansToGenerate) {
+      return super.filterUniqueSolutions(
```
kiuby88: why is it necessary `super` here?

kiuby88:
It looks like for me

planner/optimizer/optimizer-core/src/main/java/eu/seaclouds/platform/planner/optimizer/heuristics/BlindSearch.java
perezp: Thanks @kiuby88 . It is not necessary. I'll remove it. I wrote it only for clarity (to myself) about method location.

PR #271, opened 2016-03-22
Feature(Planner): PaaS support improvement in DAM generator

PaoloCifariello:
In this PR some problems related to the DAM generation in case of PaaS deployment are addressed.

@kiuby88 can you please take a look?


rosogon:
@PaoloCifariello : Are there tests of node_templates to be deployed on PaaS?

PR #274, opened 2016-03-29
Improve/second dam generator refactoring

kiuby88:
Big DamGenerator Refactoring.

PaaS support was included.
JBDC relations for IaaS and PaaS were added.
EndpointRelations for IaaS and PaaS were added.

TODO: 
* Requirements must be refactoring
* SLA should  receive Monitoring Rules
* Repairing


kiuby88:
@perezp @rosogon

deployer/src/main/java/org/apache/brooklyn/entity/cloudfoundry/webapp/PaasWebAppCloudFoundryDriver.java
```diff
@@ -65,7 +65,7 @@ private void initApplicationParameters() {
         if(!Strings.isBlank(getEntity().getConfig(CloudFoundryWebApp.APPLICATION_NAME))){
             applicationName = getEntity().getConfig(CloudFoundryWebApp.APPLICATION_NAME);
         } else {
-            applicationName = "cf-app" + Identifiers.makeRandomId(8);
+            applicationName = "cf-app-" + Identifiers.makeRandomId(8);
```
rosogon: Is there any way to use as prefix the module name?

kiuby88: Yes, CF application requires an unique id, if this id is not specify from the DAM, then the entity will generate and use this id

planner/core/src/main/java/eu/seaclouds/platform/planner/core/application/decorators/SeaCloudsManagmentPolicyDecorator.java
```diff
@@ -0,0 +1,138 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import javax.xml.bind.JAXBContext;
+import javax.xml.bind.JAXBException;
+import javax.xml.bind.Marshaller;
+import java.io.StringWriter;
+import java.util.Map;
+
+public class SeaCloudsManagmentPolicyDecorator implements ApplicationFacadeDecorator {
```
rosogon: Rename to SeaCloudsManagementPolicyDecorator

rosogon:
@kiuby88 : Nice way to implement the generator. I would like to test it live.

andreaturli:
I don't think this is a refactoring: 41 commits, 64 files changed are a completely new implementation, instead. I don't think we should merge such a huge PR at this stage of the project (March 30!) as it is too risky for the project itself. wdyt @SeaCloudsEU/developers ?

ahgittin:
an impressively huge piece of work but as a consequence a proper review will take time and high risk merging without that

kiuby88:
Thanks toy four reply @andreaturli and @ahgittin. 
Probably refactoring it is not the best work in this case because DamGenerator was rewritten completly. 
In any case, only DamGenerator was modified the interface it is the same and the produced artifacts (DAM, ADP, AAM) it also is the same. 
Probably, I should try to open little PR in order to improve the DamGenerator step by step but I develop this PR during the Easter week and I decided not open little PRs during the Easter week. Probably it was not the best decision because it is not the best team working technique.

PR integration it is not my decision but SeaClouds did not work before this PR. We have tested the results on a testing environment and it _works_, in any case I agree that a lot of refactoring could be done and **a lot of unit and integration tests** should be added.

rosogon:
In my opinion, it is a higher risk not merging this. It works for ATOS case study.

fradandria:
I would agree with Roman. Do we have alternatives? This PR resolves multiple issue regarding the ATOS use case implementation. we finally made this end to end demo deploying the ATOS services.... 
@kiuby88 Could we make additional tests, before to accept the PR? in any case I do not think here we are discussing about the utility of the new code ... dont' we?

rosogon:
We have been using this code several days. Sure there are things to fix, but it is much better than the current code. LGTM. Please, @kiuby88 , squash in more meaningful commits.

perezp:
Once we have seen the desing+plan+deployment demo using the piece of code in the PR, I agree with merging it.

I also agree with @ahgittin . This PR would require a lot of time to be properly reviewed. On the one hand, I see that it is risky to merge it as it now since it may contain bugs and create technical debt if the design of the solution is not  good enough. On the other hand, this PR does not break any piece of code that were currently working. It is a step forward from the current situation.

PR #277, opened 2016-04-04
Replicate host node templates

perezp:
This PR replicates in the ADP the node template A that is in a "host" relationship with other node templates B,C,... . 

With this change, the semantic of the node templates that are "host" of others in the ADP changes: 
- from the node template defining a type of cloud resource (e.g. Amazon EC2 m3.medium machine)
- to the node template defining a cloud resource instance. 

@kiuby88 Please, take a look as this is useful for the DAMgenerator
@PaoloCifariello  I've changed slightly the tests in planner.core package since the were asserting properties that were not valid anymore.


kiuby88:
+1
@perezp thanks a lot

PR #279, opened 2016-04-05
Fixes refresh issue in status screen

rsucasas:
This pull request fixes the refresh issues in status screen - dashboard

### DIFF
--- a/dashboard/src/main/resources/webapp/static/lib/angular-topology-editor/angular-topology-editor.js
+++ b/dashboard/src/main/resources/webapp/static/lib/angular-topology-editor/angular-topology-editor.js
@@ -118,11 +118,16 @@
                         scope.canvas = Canvas();
                         scope.canvas.init(uniqid, canvasOptions);
                         Status.init(scope.canvas);
+								Status.fromjson(scope.topology);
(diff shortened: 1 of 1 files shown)


dashboard/src/main/resources/webapp/static/lib/angular-topology-editor/angular-topology-editor.js
```diff
@@ -118,11 +118,16 @@
                         scope.canvas = Canvas();
                         scope.canvas.init(uniqid, canvasOptions);
                         Status.init(scope.canvas);
-
-                        scope.$watch('topology', function (newValue) {
-                            Status.fromjson(scope.topology);
-                            scope.canvas.restart();
-                        });
+								Status.fromjson(scope.topology);
```
rosogon: @rsucasas. Minor issue: you should convert those tabs into spaces.

rosogon: @rsucasas : Status.fromjson should be inside the $watch, and executed only when seaCloudsApplicationId changes.

PR #281, opened 2016-04-05
Feature/php env variables

kiuby88:
Adding PHP env variables on IaaS and PaaS
Adding PHP db relations from IaaS and PaaS.
Adding DAMs for nuro from Iaas and Paas


kiuby88:
@rosogon could you take a look please?

PR #284, opened 2016-04-07
Feature/php relations

kiuby88:
PHP relation management.
PHP final DAM examples (Nuro-Complex app)
Minor grafana dashboard generation issue (fixed)
Support .zip artifacts for PHP deployment on IaaS


kiuby88:
@rosogon take a look please

deployer/src/main/java/org/apache/brooklyn/entity/php/httpd/PhpHttpdSshDriver.java
```diff
@@ -236,18 +255,17 @@ private String configureHttpPort() {
     //TODO refactos using the strategy pattern
     private String installPhp() {
         log.debug("Installing PHP v", new Object[]{getEntity().getPhpVersion()});
-        if(getEntity().getPhpVersion().equals("5.4")){
+        if (getEntity().getPhpVersion().equals("5.4")) {
```
rosogon: Why only php5.4 has a different treatment?

deployer/src/test/java/eu/seaclouds/apache/PhpHttpdServerIntegrationTest.java
```diff
@@ -49,7 +49,7 @@
 
     private TestApplication app;
     private String gitRepoURLApp="https://github.com/kiuby88/phpHelloWorld.git";
-    private String tarballResourceUrl="https://github.com/kiuby88/phpHelloWorld/archive/1.tar.gz";
+    private String tarballResourceUrl="https://s3-us-west-2.amazonaws.com/seaclouds-usescases-artifacts/php/phpHelloWorld.tgz";
```
rosogon: When the amazon account expires, this will not work anymore, I think. Could this tarball be placed inside src/test/resources?

rosogon:
@kiuby88 : only some comments. I am not an expert in this, but I don't see any big error, and this have been tested deploying nuro use case since several days.

deployer/src/main/java/org/apache/brooklyn/entity/php/httpd/PhpHttpdSshDriver.java
kiuby88: By default 5.5 is installed. 5.4 requires a different installation process. Older version are not supported yet.

deployer/src/test/java/eu/seaclouds/apache/PhpHttpdServerIntegrationTest.java
kiuby88: Oops

rosogon:
@kiuby88 : +1 .Please, squash

rosogon:
Merging. Tests passed before the squash

Issue #180, opened 2015-12-04
Missing Grafana + InfluxDB installation with the installer


adriannieto:
Grafana and InfluxDB should be integrated when SeaClouds is installed

adriannieto:
We could use this as an starting point. https://github.com/adriannieto/SeaCloudsPlatform/tree/feature/grafana-integration

Issue #280, opened 2016-04-05
Credentials screen not working after new DAM format


rosogon:
PaaS providers do not appear as node templates. The information should be taken brooklyn policies.

PR #285, opened 2016-04-08
Fixes issue #180: Credentials screen not working after new DAM format

rosogon:
In new DAM Generator, PaaS providers do not appear as node templates and
the information should be taken from brooklyn policies.

When generating the topology associated to the DAM, the DAM is now
preprocessed to add PaaSes as node templates in the DAM. This only takes
effect for the generation of the topology, and allows the rest of the
code work as it was.


rosogon:
@kiuby88 , @perezp : Please, could you review?

kiuby88:
+1

rosogon:
Closes #280

PR #292, opened 2016-04-17
Adds a textarea in step 1 of wizard to specify an initial topology.

rosogon:
The topology is to be pasted inside a textarea. The textarea is
initially hidden, and is shown when a checkbox is checked.


kiuby88:
+100000
This feature must have be added some months ago!!

PR #293, opened 2016-04-17
Fixes jumping topology on status screen.

rosogon:
Status.fromjson recreates() the whole topology structure. This commit
calls Status.fromjson only when the application changes.

TODO: the new code may update other properties than status (e.g., module
urls).


kiuby88:
+1

PR #294, opened 2016-04-18
Adds a textarea to modify agreement in the last wizard step.

rosogon:
The agreement textarea is below dam textarea. Both are collapsed
initially.

The agreement is loaded from the encoded agreement in the dam.
Before deployment, the agreement (modified or not) is stored in the dam.


kiuby88:
+1

PR #296, opened 2016-04-19
Various fixes to the rules and data collectors generation for PaaS scenarios.

MicheleGuerriero:
This PR apply various fixes to the generation of monitoring information for SeaClouds applications, in particular for PaaS deployments. 

* Fixed a bug which was preventing the generation of monitoring rules for PaaS modules.
* The javaAppDc node template is not generated anymore since it is within the application and just need to be configured through the export of few environmental variables. These are now added under the "env" field of the node template representing the module which contains the javaAppDc.
* The seacloudsDc node template is not generated anymore since it is part of the SeaClouds platform and just need to be configured (not actually deployed). This configuration task is now missing and will rely on the SeaCloudsManagementPolicy in an upcoming PR.

@kiuby88 can you have a look at this?


kiuby88:
Thanks a lot @MicheleGuerriero 
+1

PR #299, opened 2016-04-25
Fixes a javascript error and improves nodes look in status screen.

rosogon:
Previously, the nodes were rounded by a green or red line (not easy to distinguish between them). Now  the background is in light green or red.


kiuby88:
It looks good
+1

PR #305, opened 2016-05-03
Updating codecov file

kiuby88:


codecov.yml
```diff
@@ -1,5 +1,20 @@
...
+
   status:
+    changes: false
+    project: false
+    #project:
+    #  target: auto
+    #  if_no_uploads: error
+
     patch: false
+    #patch:
```
stevepeak: The syntax here should be:

```yaml
coverage:
  status:
    patch:
      default:  # note this context
        target: "80%"
```

kiuby88: Thanks for your reply @stevepeak

PR #301, opened 2016-05-02
Feature/adding seaclouds dc to catalog

MicheleGuerriero:
This PR adds the deployment of the seaclouds-dc in the catalog.bom in order to be deployed within the SeaClouds Platform. At the end of the deployment of the seaclouds-dc it is also installed a monitoring rule that need to be statically added to the Monitor Manager. @kiuby88 @rosogon can you have a look?

PR #307, opened 2016-05-04
Fixes type of sla-db and updates the creation script

rosogon:
Updates the url of the creation script. The new script creates an additional user to connect from localhost.


kiuby88:
LGTM +1

I see that you have modified MySql. I have added this in #301 too

PR #309, opened 2016-06-10
Subchild support

kiuby88:
Solving a little issue regarding the children of an application that have to be monitoring.


rosogon:
+1

PR #311, opened 2016-07-13
Updating deployer and brooklyn-tosca links.

kiuby88:


kiuby88:
@rosogon could you take a look please?

rosogon:
Have you tested it? I have the impression that if the deployer jar is not renamed, it won't work. That's probably you hosted the jar in amazon.

kiuby88:
@rosogon take a look

rosogon:
+1

